PR #2, opened 2015-08-16
Some fixes to packaging

gdamjan:
make sure the sub-packages are included when installing.
add dependencies so they are automatically pulled with pip

bump the version and don't forget to upload an archive too:
```python setup.py sdist upload```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 import codecs
 import os.path
 
-from setuptools import setup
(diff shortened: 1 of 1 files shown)


horazont:
So I messed up the merge, but your commits are included with proper attribution (2136e18 and 3460632). Thank you very much for pointing out the issues!

I also uploaded the new version to pypi.

PR #5, opened 2016-03-13
deprecations fixed

sdfsdhgjkbmnmxc:
async() => ensure_future()
assertRaisesRegexp() => assertRaisesRegex()


horazont:
Unfortunately, ensure_future is not in Python 3.4, which is what I use. Could you rebase your branch so that the commit touching asyncio.async is not in there?

In any case, thank you for your pull requests.

horazont:
I rebased your commit about assertRaisesRegexp onto devel and merged it. See 95a3783.

I will use your other commit when the time has come for #11.

Issue #21, opened 2016-07-19
Add xmpp_block to examples


mcepl:
https://gitlab.com/mcepl/jabberd_utils/blob/master/xmpp_block

It is a script blocking all JIDs (full or partial) added on the command line to the [XEP-0191](https://xmpp.org/extensions/xep-0191.html) blocking list. Could be useful for combating current wave of Jabber spam.

horazont:
I included the example by re-structuring it a bit. It is in ``examples/xep191_block.py``. Please test if you can, because the servers I use do not support that feature currently. The CLI has changed slightly, but --help and the examples readme should show you the way around.

It also requires Python 3.5+ to run.

mcepl:
Unfortunately, not python3.5 around, so I just have to trust you (and note, that it seems sad you require just the very very latest release of Python).

horazont:
@mcepl Good point, aioxmpp itself is still python 3.4-compatible.

I have pushed a few commits to master which backport the examples to Python 3.4, and if there will be a 0.7.2 release, it will contain the backported examples.

Please test the block_jid example.

mcepl:
@horazont what do you think about this? That makes the script directly useable.

mcepl:
```
matej@mitmanek: ~$ block_jid --add jabber.me
Traceback (most recent call last):
  File "/home/matej/.local/bin/block_jid", line 192, in <module>
    exec_example(BlockJID())
  File "/home/matej/archiv/knihovna/repos/aioxmpp/examples/framework.py", line 168, in exec_example
    loop.run_until_complete(example.run_example())
  File "/usr/lib64/python3.4/asyncio/base_events.py", line 316, in run_until_complete
    return future.result()
  File "/usr/lib64/python3.4/asyncio/futures.py", line 275, in result
    raise self._exception
  File "/usr/lib64/python3.4/asyncio/tasks.py", line 238, in _step
    result = next(coro)
  File "/home/matej/archiv/knihovna/repos/aioxmpp/examples/framework.py", line 155, in run_example
    yield from self.run_simple_example()
  File "/home/matej/.local/bin/block_jid", line 169, in run_simple_example
    iq
  File "/home/matej/archiv/knihovna/repos/aioxmpp/aioxmpp/stream.py", line 2129, in send_iq_and_wait_for_reply
    reply = yield from fut
  File "/usr/lib64/python3.4/asyncio/futures.py", line 386, in __iter__
    yield self  # This tells Task to wait for completion.
[...]

PR #27, opened 2016-10-10
Add default configuration file for aioxmpp examples.

mcepl:
Perhaps helping #21 ?

### DIFF
--- a/examples/block_jid.py
+++ b/examples/block_jid.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 ########################################################################
 # File name: block_jid.py
 # This file is part of: aioxmpp
--- a/examples/block_jid.py
+++ b/examples/block_jid.py
@@ -184,8 +184,7 @@ class BlockJID(Example):
 
             # print all the items; again, .items is a list of JIDs
             print("current block list:")
-            for item in result.items:
-                print(" ", item)
+            print('\n'.join(sorted(['\t' + str(x) for x in result.items])))
 
 
 if __name__ == "__main__":
--- a/examples/framework.py
+++ b/examples/framework.py
@@ -116,6 +116,7 @@ class Example(metaclass=abc.ABCMeta):
             )
 
         self.g_jid = self.args.local_jid
+
         if self.g_jid is None:
             try:
                 self.g_jid = aioxmpp.JID.fromstr(
@@ -142,7 +143,15 @@ class Example(metaclass=abc.ABCMeta):
             if self.args.ask_password:
                 password = getpass.getpass()
             else:
-                password = self.config.get("global", "password")
+                try:
+                    jid_sect = str(self.g_jid)
+                    if jid_sect not in self.config:
+                        jid_sect = "global"
+                    password = self.config.get(jid_sect, "password")
+                except configparser.NoOptionError:
+                    logging.error(('When the local JID %s is set, password ' +
+                                   'must be set as well.') % str(self.g_jid))
+                    raise
         else:
             password = None
             anonymous = anonymous or ""
--- a/examples/framework.py
+++ b/examples/framework.py
@@ -27,6 +27,8 @@ import getpass
 import json
 import logging
 import logging.config
+import os
+import os.path
 import signal
 import sys
 
@@ -44,9 +46,16 @@ class Example(metaclass=abc.ABCMeta):
         self.argparse = argparse.ArgumentParser()
 
     def prepare_argparse(self):
+
+        config_default_path = os.path.join(
+            os.environ.get("XDG_CONFIG_HOME", os.path.expanduser("~/.config")),
+            "aioxmpp_examples.ini")
+        if not os.path.exists(config_default_path):
+            config_default_path = None
+
         self.argparse.add_argument(
             "-c", "--config",
-            default=None,
+            default=config_default_path,
             type=argparse.FileType("r"),
             help="Configuration file to read",
         )


examples/block_jid.py
```diff
@@ -114,6 +116,11 @@ def jid(s):
             help="If given, prints the block list at the end of the operation",
         )
 
+        config_default = os.path.expanduser(
+            '~/.config/aioxmpp_credentials.ini')
```
horazont: I would prefer using the current working directory or the directory the examples are located in. Using ~/.config is not portable (we would have to inspect XDG_CONFIG_HOME on linuxes and who-knows-what on other OSes) and using XDG or appdirs is overkill here.

examples/block_jid.py
```diff
@@ -114,6 +116,11 @@ def jid(s):
...
+        if os.path.exists(config_default):
+            self.argparse.set_defaults(config=config_default)
```
horazont: Please move the whole logic to framework.py, so that other examples can benefit from that.

examples/block_jid.py
mcepl: Can we at least settle on ``~/.aioxmpprc``? There **is** ``$HOME`` directory on Windows (yes, it is in the wrong place, etc.).

horazont: I see that you have a strong opinion towards something HOME-based. What is the reason for that? I don’t see one, but if you explain yours to me it might start to make sense to me. Currently, I do not see any advantage over simply defaulting to the examples directory (which is easy to detect, something along the lines of ``os.path.dirname(sys.argv[0])``).

mcepl: Well, many many years of experience persuaded me that mixing code (not mentioning Linux distro packages) and configuration is the path to hell. Configuration should be separate from the code, and ``$HOME`` (or ``~/.config``) is the place to put the configuration to, so why not to do it properly? I guess you are on Mac (where there is usually a way less of discipline), aren't you?

horazont: In general I agree. I don’t think it makes too much sense for these examples though, because they are not installed with neither the pip nor the AUR package. Their intended audience is not end users, but developers who want to take a look on how to use the aioxmpp API.

I can see some advantages when one has multiple aioxmpp repositories checked out, which is handy while preparing releases.

As a middle way, let’s go with:

```
os.path.join(os.environ.get("XDG_CONFIG_HOME", os.path.expanduser("~/.config")), "aioxmpp_examples.ini")
```

This will work correctly on XDG-compliant distributions, and everyone else gets a more-or-less usable default. That doesn’t hurt, because before there was no default at all, the file is unlikely to exist or contain any surprising content, and one can still override that using the command line option.

mcepl:
Something like this?

examples/framework.py
```diff
@@ -116,7 +125,10 @@ def configure(self):
         if self.args.ask_password:
             password = getpass.getpass()
         else:
-            password = self.config.get("global", "password")
+            if self.config.has_section(self.g_jid):
+                password = self.config.get(self.g_jid, 'password')
```
horazont: nice idea! take care though that g_jid is a JID instance, does that work properly with ConfigParser?

examples/framework.py
```diff
@@ -43,9 +45,16 @@ def __init__(self):
         self.argparse = argparse.ArgumentParser()
 
     def prepare_argparse(self):
+
+        config_default_path = os.path.join(os.environ.get("XDG_CONFIG_HOME",
+                                           os.path.expanduser("~/.config")),
+                                           "aioxmpp_examples.ini")
```
horazont: Break the line between ``(`` and ``os.environ`` please, that will allow to align the default for ``.get`` with the opening parenthesis, making the line easier to read.

examples/framework.py
mcepl: I don’t understand. As long as ``aioxmpp.structs.JID`` has ``__str__`` then it should not be a problem, should it? Or what else do you see there?

mcepl:
better?

examples/framework.py
horazont:
```
horazont@sinistra tmp › cat foo.ini
[foo@bar.baz]
test=test

horazont@sinistra tmp › python3
Python 3.5.2+ (default, Sep 22 2016, 12:18:14) 
[GCC 6.2.0 20160927] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import configparser
>>> p = configparser.ConfigParser()
>>> p.read("foo.ini")
['foo.ini']
>>> p.has_section("foo@bar.baz")
True
>>> import aioxmpp
>>> j = aioxmpp.JID.fromstr("foo@bar.baz")
>>> str(j) == "foo@bar.baz"
True
>>> p.has_section(j)
False
>>> p.has_section(str(j))
True
>>> 

```

It appears that has_section will always return false when passed a JID instance.

mcepl:
right. fixed.

horazont:
Thanks!

This will go into 0.8.

examples/block_jid.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/python3
```
SamWhited: This will break in virtualenv's; if this is necessary set it to `/usr/bin/env python3` to ensure the correct version (global or the one in the virtualenv) is used.

mcepl:
right

examples/block_jid.py
horazont: Thanks! So now I know why people do that!

mcepl:
You still haven’t closed this pull request, so I will misused it for one more nitpick (script crashed with missing password).

horazont:
It isn’t supposed to crash, but to attempt to use SASL ANONYMOUS. Please give additional information and make sure you have aiosasl 0.3 or newer installed.

mcepl:
> It isn’t supposed to crash, but to attempt to use SASL ANONYMOUS.

With this patch against my latest commits:
```diff
diff --git a/examples/framework.py b/examples/framework.py
index e026087..93e2354 100644
--- a/examples/framework.py
+++ b/examples/framework.py
@@ -147,8 +147,9 @@ class Example(metaclass=abc.ABCMeta):
             if self.config.has_section(str(self.g_jid)):
                 password = self.config.get(str(self.g_jid), 'password')
             else:
-                password = self.config.get("global", "password",
-                                           fallback=None)
+                # password = self.config.get("global", "password",
+                #                            fallback=None)
+                password = self.config.get("global", "password")
             anonymous = anonymous or ""
 
         no_verify = self.config.getboolean(
```
if I miss global password and there is no section for ``matej@ceplovi.cz`` I get this:
```python
$ block_jid -j matej@ceplovi.cz -l
  File "/usr/lib64/python3.4/configparser.py", line 762, in get
    value = d[option]
  File "/usr/lib64/python3.4/collections/__init__.py", line 814, in __getitem__
    return self.__missing__(key)            # support subclasses that define __missing__
  File "/usr/lib64/python3.4/collections/__init__.py", line 806, in __missing__
    raise KeyError(key)
KeyError: 'password'

During handling of the above exception, another exception occurred:

  File "/home/matej/.local/bin/block_jid", line 191, in <module>
    exec_example(BlockJID())
  File "/home/matej/archiv/knihovna/repos/aioxmpp/examples/framework.py", line 220, in exec_example
    example.configure()
  File "/home/matej/.local/bin/block_jid", line 129, in configure
    super().configure()
  File "/home/matej/archiv/knihovna/repos/aioxmpp/examples/framework.py", line 145, in configure
    password = self.config.get("global", "password")
  File "/usr/lib64/python3.4/configparser.py", line 765, in get
    raise NoOptionError(option, section)
configparser.NoOptionError: No option 'password' in section: 'global'
$
```
``configparser.ConfigParser.get()`` doesn’t have automatic default value for the missing key.

horazont:
I think you made a mistake when rebasing/merging your changes onto/into latest devel. That branch of the ``if`` isn’t supposed to do anything with a password. If -A is given (args.anonymous is not False), password is supposed to be set to None.

I think you intended to put the part with ``str(self.g_jid)`` in the other branch.

mcepl:
OK, I have restored my branch according to your instructions, but now I am getting this again:
```python
$ block_jid -j matej@ceplovi.cz -l
  File "/usr/lib64/python3.4/configparser.py", line 762, in get
    value = d[option]
  File "/usr/lib64/python3.4/collections/__init__.py", line 814, in __getitem__
    return self.__missing__(key)            # support subclasses that define __missing__
  File "/usr/lib64/python3.4/collections/__init__.py", line 806, in __missing__
    raise KeyError(key)
KeyError: 'password'

During handling of the above exception, another exception occurred:

  File "/home/matej/.local/bin/block_jid", line 191, in <module>
    exec_example(BlockJID())
  File "/home/matej/archiv/knihovna/repos/aioxmpp/examples/framework.py", line 215, in exec_example
    example.configure()
  File "/home/matej/.local/bin/block_jid", line 129, in configure
    super().configure()
  File "/home/matej/archiv/knihovna/repos/aioxmpp/examples/framework.py", line 145, in configure
    password = self.config.get("global", "password")
  File "/usr/lib64/python3.4/configparser.py", line 765, in get
    raise NoOptionError(option, section)
configparser.NoOptionError: No option 'password' in section: 'global'
$
```

mcepl:
And of course, I get to the line 149, because both ``anonymous`` and ``self.args.ask_password`` are ``False``. So, we have to add some protection against missing password there. Either we have to stop here, because we cannot work without the password, or we will pass something (``None``?) further down the road, and let the program handle the situation

horazont:
True, this could use better handling (probably by making the password behave like other pieces of configuration and ask on the terminal if the config option is missing). But IMO, that’s out of scope for this pull request.

Also, you seem to have removed the part with configuring the password per-jid accidentally.

mcepl:
> Also, you seem to have removed the part with configuring the password per-jid accidentally.

Yes, I did, because it was truly failed rebase or a wrong way how to handle this situation. So, what would you prefer (I don't like spread around known-to-be-broken code), to raise an exception or to continue with ``None``?

horazont:
I find the NoOptionError sufficient. These examples are not meant as end user products anyways. They are aimed at developers learning the API and don’t need to be fully polished with regards to parametrisation. A developer will understand what configparser.NoOptionError is about.

mcepl:
Would be this enough? At least some explanation is provided.

PR #54, opened 2017-03-25
Implementation of XEP-0084 User Avatar

sebastianriese:


aioxmpp/avatar/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+Data Representation
+===================
+
+The following class is used to describe the possible locations of an
+avatar image:
+
+.. autoclass:: AvatarSet
+
+.. autoclass:: aioxmpp.avatar.service.AbstractAvatarDescriptor
```
horazont: Use something like:
```
.. module:: aioxmpp.avatar.service
.. currentmodule:: aioxmpp.avatar.service
.. autoclass:: AbstractAvatarDescriptor
```

instead. This can be found in other service docs too, grep for 'module::'

aioxmpp/avatar/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+"""
+
+from .service import (AvatarSet, AvatarServer, AvatarClient)
```
horazont: Add ``# NOQA`` to the end of line (if you need to break the line afterwards, the ``# NOQA`` must be on the first line) to prevent flake8 complaints.

aioxmpp/avatar/service.py
```diff
@@ -0,0 +1,536 @@
...
+
+    def __init__(self, client, **kwargs):
+        super().__init__(client, **kwargs)
+        self._metadata_cache = {}
+
+        self._pubsub = self.dependencies[pubsub.PubSubClient]
+
+    def _cook_metadata(self, jid, metadata):
+        result = collections.defaultdict(lambda: [])
+        for metadata_item in metadata.payload.items:
```
horazont: This loop feels too deeply nested, but I also can’t find a better way to word it.

aioxmpp/avatar/service.py
```diff
@@ -0,0 +1,536 @@
...
+        namespaces.xep0084_metadata + "+notify"
+    )
+
+    on_metadata_changed = callbacks.Signal(doc=
+    """
+    Fires when avatar metadata changes.
+
+    :param jid: The JID which the avatar belongs to.
+    :param metadata: The list of metadata descriptors.
+    """)
```
horazont: Fire with ``# NOQA`` at this until the warnings are gone *or* put it inline into the class docstring with:

```
.. signal:: on_metadata_changed(…)
```

At the very least, you need to add the arguments to the ``.. autosignal::``, because they cannot be inferred from the signal declaration.

tests/avatar/test_service.py
```diff
@@ -0,0 +1,630 @@
...
+#
+########################################################################
+import base64
+import binascii
+import hashlib
+import unittest
+
+import aioxmpp
+import aioxmpp.errors as errors
+import aioxmpp.xso.model as xso
```
horazont: Please do not import ``aioxmpp.xso.model`` directly, use ``aioxmpp.xso``. It should export everything you need.

horazont:
Stupid interaction with gitolite, sorry for deleting :-)

sebastianriese:
Just discovered a bug: Two concurrent calls for publish_avatar_set or get_avatar_metadata can lead to a race condition. The services defined by avatar should have a lock to prevent this. In the get_avatar_metadata case this can cause extra network traffic, in the publish_avatar_set case it may cause inconsistent state in pubsub (with the metadata node not matching the data node).

horazont:
A lock won’t fix this for ``publish_avatar_set``: The same race may occur between two clients.

We need something more sophisticated. Off the top of my head I would suggest the following: check whether data and metadata are consistent after the write operation (note: I write "consistent", not "equal to what we wrote") and if not, wait for a random delay (between 0.5 and 2 seconds maybe?) and re-try the write (maybe doubling the random range edges each time?). Impose a limit on the maximum number of writes attempted (maybe 3?) and if it isn’t consistent afterwards, raise an exception. Log when retries are needed.

Of course, we can *in addition* have a lock to prevent this situation between local users of the service.

aioxmpp/avatar/__init__.py
```diff
@@ -50,8 +50,11 @@
 
 .. autoclass:: AvatarSet
 
-.. autoclass:: aioxmpp.avatar.service.AbstractAvatarDescriptor
+.. module:: aioxmpp.avatar.service
+.. currentmodule:: aioxmpp.avatar.service
+.. autoclass:: AbstractAvatarDescriptor
+   :members:
```
horazont: Don’t use ``:members:``. Document the members explicitly using ``.. automethod`` et al. This allows for grouping the members according to their use-cases and adding explanatory text in-between, as well as excluding members which are not supposed to show up in the docs.

(on the downside, you can also easily forget to list a member ... :))

aioxmpp/avatar/service.py
```diff
@@ -301,12 +301,19 @@ class AvatarClient(service.Service):
...
-    .. autosignal:: on_metadata_changed
+    .. note:: :class:`AvatarClient` only caches the metadata, not the
+              actual image data. This is the job of the caller.
+
+    .. signal:: on_metadata_changed(jid, metadata)
+
+        Fires when avatar metadata changes.
+
+        :param jid: The JID which the avatar belongs to.
+        :param metadata: The list of metadata descriptors.
```
horazont: Add type information to the parameters, e.g. ``:type metadata: :class:`something` ``.

aioxmpp/avatar/service.py
```diff
@@ -337,35 +338,37 @@ def __init__(self, client, **kwargs):
         self._pubsub = self.dependencies[pubsub.PubSubClient]
 
     def _cook_metadata(self, jid, metadata):
+        def iter_metadata_info_nodes(metdata):
+            for item in metadata.payload.items:
+                info_map = item.registered_payload.info
+                for mime_type in info_map:
+                    for metadata_info_node in info_map[mime_type]:
+                        yield metadata_info_node
```
horazont: The code looks much better this way!

Suggestion: You could refactor the last three lines into a property/method of the avatar metadata XSO.

aioxmpp/pubsub/service.py
```diff
@@ -537,7 +537,8 @@ def publish(self, jid, node, payload, *, id_=None):
         the node, it is replaced. If no ID is given, a ID is generated by the
         server.
 
-        Return the ID of the item as published.
+        Return the ID of the item as published (or :data:`None` if the server
+        does not inform us).
```
horazont: While you’re touching this docstring, can you please update it to use ``:param:`` and ``:return:`` and so on in addition to the explanatory text?

examples/retrieve_avatar.py
```diff
@@ -0,0 +1,91 @@
...
+            "--remote-jid",
+            type=jid,
+            help="the jid of which to retrieve the avatar"
+        )
+
+    def configure(self):
+        super().configure()
+
+        self.output_file = self.args.output_file
+        self.remote_jid = self.args.remote_jid
```
horazont: This will raise AttributeError if ``--remote-jid`` has not been passed; argparse does not have implicit defaults, you have to set a default explicitly in the ``add_argument`` call.

examples/retrieve_avatar.py
```diff
@@ -0,0 +1,91 @@
...
+        if self.remote_jid is None:
+            try:
+                self.remote_jid = aioxmpp.JID.fromstr(
+                    self.config.get("avatar", "remote_jid")
```
horazont: Document the setting in the ``examples`` README please.

aioxmpp/avatar/service.py
```diff
@@ -0,0 +1,539 @@
...
+        This will raise :class:`NotImplementedError` if we are not
+        capable to retrieve the image data. It is guaranteed to not
+        raise :class:`NotImplementedError` if :attr:`image_in_pubsub`
+        is true.
+        """
+        raise NotImplementedError
+        yield
+
+    @property
+    def image_in_pubsub(self):
```
horazont: Rename to ``is_image_in_pubsub`` or ``has_data_in_pubsub`` or something like that. It’s not clear that this is a boolean flag otherwise when reading code using it (I stumbled across this when reading the examples).

examples/set_avatar.py
```diff
@@ -0,0 +1,74 @@
...
+            "avatar_file",
+            help="the file PNG image that will be used as avatar."
+        )
+
+        self.argparse.add_argument(
+            "--wipe-avatar",
+            action="store_true",
+            default=False,
+            help="set the avatar to no avatar when exiting."
+        )
```
horazont: I feel ``--wipe-avatar`` and setting the avatar should be separate tools *or* at least separate options. It shouldn’t be doing both things in one transaction. If I want to wipe my avatar, I shouldn’t need to find an image which I can use as "last avatar before I erase my avatar".

tests/pubsub/test_service.py
```diff
@@ -1226,6 +1226,108 @@ def test_publish_without_payload(self):
...
+        self.assertIsNone(request_iq.to)
+        self.assertIsInstance(request_iq.payload, pubsub_xso.Request)
+
+        request = request_iq.payload
+        self.assertIsInstance(request.payload, pubsub_xso.Publish)
+        self.assertEqual(request.payload.node, "foo")
+        self.assertIs(request.payload.item.registered_payload, payload)
+
+
+        self.assertEqual(result, "some-other-id")
```
horazont: Too much whitespace around this one. Doesn’t flake8 complain?

examples/retrieve_avatar.py
sebastianriese: It asks for remote_jid on the terminal if none is given as argument.

horazont: Have you tested that?

sebastianriese: Yes.

horazont: Surprised, but fine :-)

examples/retrieve_avatar.py
sebastianriese: Done.

examples/set_avatar.py
sebastianriese: Done. The tool now uses the two command `--wipe-avatar` and `--set-avatar`.

aioxmpp/pubsub/service.py
sebastianriese: I would propose to do that in another PR and update *all* the docstrings, so that the documentation of the pubsub module has a consistent style.

PR #55, opened 2017-03-27
Fix the toposort of aioxmpp.Service subclasses

sebastianriese:
* fixes a aioxmpp.Client.summon bug, where services where loaded
  in the wrong order

* makes comparison of the aioxmpp.Service subclasses a topological order
  consistent with the dependency order of the services (note: this order
  can change radically when a new service is declared, the relative
  order of independent services is not defined!)

### DIFF
--- a/aioxmpp/service.py
+++ b/aioxmpp/service.py
@@ -307,7 +307,6 @@ class DependencyGraph:
             marked.remove(node)
             sorted_.append(node)
 
-        # self is not yet in the class list
         for node in self._nodes:
             visit(node)
 
--- a/aioxmpp/service.py
+++ b/aioxmpp/service.py
@@ -44,11 +44,6 @@ These decorators provide special functionality when used on methods of
 
 .. note::
 
-   Inheritance from classes which have any of these decorators on any of its
-   methods is forbidden currently, because of the ambiguities which arise.
-
-.. note::
-
    These decorators work only on methods declared on :class:`Service`
    subclasses, as their functionality are implemented in cooperation with the
    :class:`Meta` metaclass and :class:`Service` itself.
@@ -136,7 +131,7 @@ service instance using the descriptor, including cleanup.
 Metaclass
 =========
 
-.. autoclass:: Meta([inherit_dependencies=True])
+.. autoclass:: Meta()
 
 
 """
@@ -263,14 +258,73 @@ class Descriptor(metaclass=abc.ABCMeta):
         return obj
 
 
+class DependencyGraphNode:
+    def __init__(self):
+        self.class_ = None
+
+    def __repr__(self):
+        return "DependencyGraphNode({!r})".format(self.class_)
+
+
+class DependencyGraph:
+
+    def __init__(self, edges=None, nodes=None):
+        if edges is None:
+            edges = []
+        if nodes is None:
+            nodes = []
+        self._edges = set(edges)
+        self._nodes = set(nodes)
+
+    def copy(self):
+        return DependencyGraph(self._edges, self._nodes)
+
+    def add_node(self, node):
+        self._nodes.add(node)
+
+    def add_edge(self, from_, to):
+        self._edges.add((from_, to))
+
+    def toposort(self):
+        edges = collections.defaultdict(lambda: set())
+        for from_, to in self._edges:
+            edges[from_].add(to)
+
+        sorted_ = []
+        marked = set()
+        done = set()
+
+        def visit(node):
+            if node in marked:
+                raise ValueError("dependency loop in service definitions")
+            if node in done:
+                return
+            done.add(node)
+            marked.add(node)
+            for dep in edges[node]:
+                visit(dep)
+            marked.remove(node)
+            sorted_.append(node)
+
+        # self is not yet in the class list
+        for node in self._nodes:
+            visit(node)
+
+        return sorted_
+
+
 class Meta(abc.ABCMeta):
     """
     The metaclass for services. The :class:`Service` class uses it and in
     general you should just inherit from :class:`Service` and define the
     dependency attributes as needed.
 
+    Only use :class:`Meta` explicitely if you know what you are doing,
+    and you most likely do not. :class:`Meta` is internal API and may
+    change at any point.
+
     Services have dependencies. A :class:`Meta` instance (i.e. a service class)
-    can declare dependencies using the following two attributes.
+    can declare dependencies using the following attributes.
 
     .. attribute:: ORDER_BEFORE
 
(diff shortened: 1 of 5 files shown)


aioxmpp/service.py
```diff
@@ -263,14 +258,73 @@ def __get__(self, instance, owner):
         return obj
 
 
+class DependencyGraphNode:
+    def __init__(self):
+        self.class_ = None
+
+    def __repr__(self):
+        return "DependencyGraphNode({!r})".format(self.class_)
```
horazont: Do I understand it correctly that you introduced this indirection to deal with classes which are currently being constructed (and not constructed yet)? Otherwise I’d suggest to remove this class.

aioxmpp/service.py
```diff
@@ -263,14 +258,73 @@ def __get__(self, instance, owner):
...
+    def __init__(self, edges=None, nodes=None):
+        if edges is None:
+            edges = []
+        if nodes is None:
+            nodes = []
+        self._edges = set(edges)
+        self._nodes = set(nodes)
+
+    def copy(self):
+        return DependencyGraph(self._edges, self._nodes)
```
horazont: Re-implement ``__copy__`` and use ``copy.copy`` instead. Standard mechanisms and such. Although I’m not sure if this doesn’t rather qualify for a ``__deepcopy__`` because it also copies the sets.

aioxmpp/service.py
```diff
@@ -319,16 +373,32 @@ class or instance.
...
+       in :attr:`ORDER_BEFORE`.
 
-    After a class has been instanciated, the full set of dependencies is
-    provided in the attributes, including all transitive relationships. These
-    attributes are updated when new classes are declared.
+       This is primarily used internally to handle :attr:`ORDER_BEFORE` when
+       summoning services.
+
+       It is an error to manually define :attr:`PATCHED_ORDER_AFTER` in a class
+       definition, doing to will raise a :class:`TypeError`.
```
horazont: :+1:

aioxmpp/service.py
```diff
@@ -319,16 +373,32 @@ class or instance.
...
+
+    The :attr:`ORDER_AFTER` and :attr:`ORDER_BEFORE` attribtes do not
+    change after class creation (in earlier versions they contained
+    the transitive completion of the dependency relation).
```
horazont: Wrap this in a ``.. versionchanged::``

aioxmpp/service.py
```diff
@@ -558,6 +608,10 @@ class Service(metaclass=Meta):
     :meth:`aioxmpp.Client.summon` works), there is no need for you
     to introduce custom arguments, and thus there should be no conflicts.
 
+    .. note::
+
+       Inheritance from classes which subclass Service is forbidden.
```
horazont: Add a ``.. versionchanged::`` about that right below this note.

tests/test_node.py
```diff
@@ -2841,17 +2841,17 @@ def __init__(self, *args, **kwargs):
                     ),
                     dependencies={
                         Svc2: unittest.mock.ANY,
-                        Svc3: unittest.mock.ANY,
+                        # Svc3: unittest.mock.ANY,
```
horazont: Remove the comments if you don’t plan to re-enable that code later.

tests/test_node.py
```diff
@@ -2841,17 +2841,17 @@ def __init__(self, *args, **kwargs):
...
         )
 
-        self.assertIs(
-            svc1.dependencies[Svc3],
-            svc2.dependencies[Svc3],
-        )
+        # self.assertIs(
+        #     svc1.dependencies[Svc3],
+        #     svc2.dependencies[Svc3],
+        # )
```
horazont: See above.

tests/test_service.py
```diff
@@ -107,30 +116,7 @@ class Bar(metaclass=service.Meta):
...
-        )
-        self.assertSetEqual(
-            set(),
-            Foo.ORDER_BEFORE
-        )
-        self.assertSetEqual(
-            set(),
-            Baz.ORDER_AFTER
-        )
+        self.assertTrue(Foo >= Bar >= Baz)
```
horazont: I prefer to avoid ``assertTrue`` whenever possible (e.g. there is no ``assertIsSubclass(…)``, so I have to use ``assertTrue(issubclass(…))``), because the error messages of the specific methods are more explicit.

If you don’t have a strong rationale here for doing otherwise, please use ``assertGreaterEqual`` instead.

tests/test_service.py
```diff
@@ -142,30 +128,7 @@ class Bar(metaclass=service.Meta):
...
-        )
-        self.assertSetEqual(
-            set(),
-            Foo.ORDER_AFTER
-        )
-        self.assertSetEqual(
-            set(),
-            Baz.ORDER_BEFORE
-        )
+        self.assertTrue(Foo <= Bar <= Baz)
```
horazont: Same as above but with ``assertLessEqual``.

tests/test_service.py
```diff
@@ -176,7 +139,7 @@ class Bar(metaclass=service.Meta):
 
         with self.assertRaisesRegex(
                 ValueError,
-                "dependency loop: Fnord loops through .*\.(Foo|Bar)"):
+                "dependency loop in service definitions"):
```
horazont: Ah, pity that we loose the information where the loop goes. That’ll make debugging loops harder. Can we get this back without too much hassle? If it’s too much trouble, don’t bother.

tests/test_service.py
```diff
@@ -222,50 +185,33 @@ class Fourth(metaclass=service.Meta):
...
+        class A(metaclass=service.Meta):
+            pass
 
-        self.assertSequenceEqual(
-            [Fourth, Baz, Bar, Foo],
-            services
-        )
+        self.assertTrue(A <= A)
+        self.assertFalse(A < A)
+        self.assertFalse(A > A)
```
horazont: Please use the appropriate test functions here, too.

tests/test_service.py
```diff
@@ -372,24 +236,10 @@ class A(metaclass=service.Meta):
...
-            Foo.ORDER_BEFORE
-        )
-        self.assertSetEqual(
-            {Foo, A, Bar},
-            B.ORDER_AFTER
-        )
-        self.assertSetEqual(
-            set(),
-            B.ORDER_BEFORE
+        self.assertTrue(Foo > A > Bar)
```
horazont: Please use the appropriate test function here, too.

aioxmpp/service.py
sebastianriese: No, it is not deepcopy: The DependencyGraphNode objects are not copied, will use __copy__ and copy.copy.

horazont: I know that it’s not a full deepcopy. But not doing a full deepcopy is exactly those cases where you’d normally overwrite ``__deepcopy__`` (e.g. where you have references to other things, for example classes, you don’t want to duplicate).

aioxmpp/service.py
horazont: Ah, and add :class:`Service` markup here.

aioxmpp/service.py
sebastianriese: With which version number?

horazont: 0.9

sebastianriese:
The issues raised above should be addressed. Still missing: Which classes are in the service loop. Also, we could optimize the calculation of the toposort in cases where only ORDER_BEFORE or ORDER_AFTER are used (since then we can simply order the class before or after all classes defined so far). Only use of ORDER_BEFORE *and* ORDER_AFTER in a class definition can trigger a loop. I think, however, that this optimization should be handled in a new PR.

aioxmpp/service.py
```diff
@@ -319,16 +373,34 @@ class or instance.
...
+       in :attr:`ORDER_BEFORE`.
+
+       This is primarily used internally to handle :attr:`ORDER_BEFORE` when
+       summoning services.
 
-    After a class has been instanciated, the full set of dependencies is
-    provided in the attributes, including all transitive relationships. These
-    attributes are updated when new classes are declared.
+       It is an error to manually define :attr:`PATCHED_ORDER_AFTER` in a class
+       definition, doing to will raise a :class:`TypeError`.
```
horazont: This could use a ``.. versionadded:: 0.9``.

aioxmpp/service.py
```diff
@@ -347,45 +419,10 @@ class Fourth(metaclass=service.Meta):
...
-                    classes.update(getattr(base, attr))
-        classes.update(
-            mcls.transitive_collect(
-                list(classes),
-                attr,
-                set())
-        )
-        return classes
+    dependency_graph = DependencyGraph()
+    service_order = {}
```
horazont: Is there any reason not to prefix those with ``__``? They will be visible on Service classes after all.

sebastianriese: I'll have to check, that it does not break the code. And mentioning it: The `_depgraph_node` member of the service classes should be documented (and perhaps uppercased, to fit with the other autogenerated memebers).

sebastianriese: `_DEPGRAPH_NODE` is documented now.

PR #57, opened 2017-03-28
Implement XEP-0191 Blocking command

sebastianriese:


aioxmpp/blocking/service.py
```diff
@@ -0,0 +1,225 @@
...
+    This service maintains the list of blocked JIDs and allows
+    manipulating the blocklist.
+
+    Attribute:
+
+    .. autoattribute:: blocklist
+
+    Signals:
+
+    .. signal:: on_initial_blocklist_reveived(blocklist)
```
horazont: Typo (reveived vs. received)

aioxmpp/blocking/service.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    ORDER_AFTER = [aioxmpp.DiscoClient]
+
+    def __init__(self, client, **kwargs):
+        super().__init__(client, **kwargs)
+        self._blocklist = None
+        self._lock = asyncio.Lock()
+        self._disco = self.dependencies[aioxmpp.DiscoClient]
+
+        self._bse_token = client.before_stream_established.connect(
```
horazont: Uh. We need depsignal to work on Client too, I see. Will do that at some point and then we can fix this, that’s out of scope for this PR.

aioxmpp/blocking/service.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+
+        if namespaces.xep0191 not in server_info.features:
+            raise RuntimeError("server does not support blocklists!")
+
+    @asyncio.coroutine
+    def _get_initial_blocklist(self):
+        yield from self._check_for_blocking()
+
+        if self._blocklist is None:
```
horazont:
```python
if self._blocklist is not None:
    return
```

Although I’m not sure when this can actually happen. Can it happen?

aioxmpp/blocking/service.py
```diff
@@ -0,0 +1,225 @@
...
+    def _check_for_blocking(self):
+        server_info = yield from self._disco.query_info(
+            self.client.local_jid.replace(
+                resource=None,
+                localpart=None,
+            )
+        )
+
+        if namespaces.xep0191 not in server_info.features:
+            raise RuntimeError("server does not support blocklists!")
```
horazont: Force the blocklist to None when this check fails.

aioxmpp/blocking/service.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        Remove the JIDs in the sequence `jids_to_block` from the
+        client's blocklist.
+        """
+        yield from self._check_for_blocking()
+
+        if not jids_to_unblock:
+            return
+
+        cmd = blocking_xso.UnblockCommand()
```
horazont: I like to have constructors for such XSOs which make the handling easier. Those constructors need tests then.

aioxmpp/blocking/service.py
```diff
@@ -0,0 +1,225 @@
...
+                # blocklist
+                return
+
+            if (block_command.from_ is None or
+                    block_command.from_ == self.client.local_jid.bare()):
+                diff = frozenset(block_command.payload.items)
+                self._blocklist |= diff
+                must_signal = True
+
+        if must_signal:
```
horazont: Instead of having this boolean, why not set diff to ``()`` at the top and check for ``if diff:``?

aioxmpp/blocking/service.py
```diff
@@ -0,0 +1,225 @@
...
+            if (unblock_command.from_ is None or
+                    unblock_command.from_ == self.client.local_jid.bare()):
+                if not unblock_command.payload.items:
+                    diff = frozenset(self._blocklist)
+                    self._blocklist = frozenset()
+                else:
+                    diff = frozenset(unblock_command.payload.items)
+                    self._blocklist -= diff
+                must_signal = True
+        if must_signal:
```
horazont: Same here for the boolean.

horazont:
I accidentally hit approve even though at least the doc typo should be fixed, can you take care of it?

PR #63, opened 2017-03-29
Further avatar fixes and improvements

### LINKED ISSUE #59: [XEP-0084] Allow access to the original id (sha-1 of the image) to prevent issues with case sensitivity

sebastianriese:
* fix the set_avatar.py example
* some refactoring
* fix some documentation issues
* fix style issues
* fix `get_avatar_metdata(..., require_fresh=True)`
* fixes #59

### DIFF
--- a/aioxmpp/avatar/service.py
+++ b/aioxmpp/avatar/service.py
@@ -209,7 +209,6 @@ class AbstractAvatarDescriptor:
         self._remote_jid = remote_jid
         self._mime_type = mime_type
         self._id = id_
-        self._normalized_id = normalize_id(self._id)
         self._nbytes = nbytes
         self._width = width
         self._height = height
@@ -294,7 +293,7 @@ class AbstractAvatarDescriptor:
 
         This is supposed to be used for caching.
         """
-        return self._normalized_id
+        return normalize_id(self._id)
 
     @property
     def mime_type(self):
--- a/aioxmpp/avatar/service.py
+++ b/aioxmpp/avatar/service.py
@@ -228,7 +228,6 @@ class AbstractAvatarDescriptor:
         is true.
         """
         raise NotImplementedError
-        yield
 
     @property
     def has_image_data_in_pubsub(self):
@@ -339,7 +338,6 @@ class HttpAvatarDescriptor(AbstractAvatarDescriptor):
         May raise NotImplementedError
         """
         raise NotImplementedError
-        yield
 
 
 class AvatarClient(service.Service):
--- a/aioxmpp/avatar/service.py
+++ b/aioxmpp/avatar/service.py
@@ -366,7 +366,6 @@ class AvatarClient(service.Service):
             for item in metadata.payload.items:
                 yield from item.registered_payload.iter_info_nodes()
 
-
         result = collections.defaultdict(lambda: [])
         for info_node in iter_metadata_info_nodes(metadata):
             if info_node.url is not None:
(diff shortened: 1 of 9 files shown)


aioxmpp/avatar/service.py
```diff
@@ -167,13 +179,37 @@ def add_avatar_image(self, mime_type, *, id_=None,
...
+
+    .. autocoroutine:: get_image_bytes
     """
 
     def __init__(self, remote_jid, mime_type, id_, nbytes, width=None,
                  height=None, url=None, pubsub=None):
         self._remote_jid = remote_jid
         self._mime_type = mime_type
         self._id = id_
+        self._normalized_id = normalize_id(self._id)
```
horazont: Handle it in the property function. If it is a performance concern at some point, we can still fix it.

aioxmpp/avatar/service.py
```diff
@@ -401,35 +447,40 @@ def get_avatar_metadata(self, jid, *, require_fresh=False):
...
+                return self._metadata_cache[jid]
+            except KeyError:
+                pass
 
-        try:
-            return self._metadata_cache[jid]
-        except KeyError:
-            pass
+        with (yield from self._lock):
+            if jid in self._metadata_cache:
```
horazont: I feel this can be written down nicer with try. Maybe:
```python
if require_fresh:
    self._metadata_cache.pop(jid, None)
else:
    try:
        return self._metadata_cache[jid]
    except KeyError:
        pass
```

aioxmpp/avatar/service.py
```diff
@@ -509,19 +566,20 @@ def publish_avatar_set(self, avatar_set):
 
         id_ = avatar_set.png_id
 
-        yield from self._pubsub.publish(
-            None,
-            namespaces.xep0084_data,
-            avatar_xso.Data(avatar_set.image_bytes),
-            id_=id_
-        )
+        with (yield from self._lock):
```
horazont: This still lacks the race condition fix if we race with other clients, right? Do you plan to do this in this PR?

sebastianriese: Nope. This will probably need a new pull request an possible additional support from PubSub (but I'll have to read the PubSub spec to be sure). As I do not want to download the image data to verify it matches, but simply check whether the image data pubsub item has the appropriate id.

horazont: Actually, I wonder whether we already get a push notification with the avatar data or at least the item ID because PEP. Can you investigate that? Does that even help us? (I’m not sure about that.)

Downloading the full avatar does seem like not a very bad thing to do here, unless we find a saner way. Matching the metadata won’t be sufficient to detect more than half of the collision scenarios.

sebastianriese: [XEP-0163 section 3]( https://xmpp.org/extensions/xep-0163.html#publish) is a bit unclear on whether we always get a notification:
> If the publication logic dictates that event notifications shall be sent, the account owner's server generates notifications and sends them to all appropriate entities as described in the Receiving Event Notifications section of this document, as well as to any of the account owner's available resources.

If we would get a notification about the data, it would be easy to check consistency. If the server is clever enough to not send the payload along (but only notify about the ID), this would be the desired behaviour.

aioxmpp/avatar/service.py
sebastianriese: Hmm, pop of course has the notion of *using* the value we pop off the dict, instead of just discarding it. I think the explicit test is the right way to go here (and shorter!).

horazont: Okay, if you prefer it that way it’s fine.

tests/avatar/test_service.py
```diff
@@ -186,22 +196,61 @@ def test_png_id_is_normalized(self):
         )
 
     def test_error_checking(self):
```
horazont: I prefer to have those tests (a) in individual test case methods and (b) using assertRaisesRegex to at least roughly check you are catching the error you are expecting to catch.

* (a) brings the advantage of overview if a single condition is violated and sense of graveness if the whole method is borked
* (b) helps with testing that it’s not a "raise RuntimeError()" at the beginning but actual different code paths being taken

tests/avatar/test_service.py
```diff
@@ -419,6 +489,19 @@ def test_get_image_bytes(self):
             )
             self.assertEqual(res, TEST_IMAGE)
 
+    def test_HttpAvatarDescriptr(self):
```
horazont: Typo in method name (``Descriptor`` vs. ``Descriptr``)

horazont:
Github claims one of my comments is outdated, but I don’t think it in fact is.

Issue #61, opened 2017-03-29
Add option to xso.Child descriptor for strict type checking and use it for PubSub Items and IQs


horazont:
It is annoying that currently, if you simply forget to designate your class as payload for IQ stanzas or PubSub Items, you are able to *send* data, but from the application perspective you only get an exception (IQ stanzas) or empty item (PubSub) back.

It would be better if that would fail early, e.g. at assignment time to that descriptor. We should make the typechecking optional, because ``isinstance`` is rather expensive and in most of the cases, XSOs are an implementation detail. This is obviously different for IQ stanzas and PubSub Items, where a friendlier (in the sense of failing loud and early when you make a mistake) interface is appropriate.

Adding a ``strict`` flag to the xso.Child descriptor or even better building on top of the validator or typing interface seems reasonable.

PR #64, opened 2017-03-29
Implement Message Carbons client (and two dependent XEPs)

### LINKED ISSUE #42: XEP-0280 (Message Carbons) support
[XEP-0280 (Message Carbons)](https://xmpp.org/extensions/xep-0280.html) and required fixes for MUC

horazont:
This would fix #42 and also implement [XEP-0203 (Delayed Delivery)](https://xmpp.org/extensions/xep-0203.html#schema) and [XEP-0297 (Stanza Forwarding)](https://xmpp.org/extensions/xep-0297.html).

### DIFF
--- a/aioxmpp/carbons/xso.py
+++ b/aioxmpp/carbons/xso.py
@@ -23,6 +23,9 @@ import aioxmpp.xso as xso
 
 from aioxmpp.utils import namespaces
 
+from ..misc import Forwarded
+from ..stanza import Message
+
 
 namespaces.xep0280_carbons_2 = "urn:xmpp:carbons:2"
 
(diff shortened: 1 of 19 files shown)


aioxmpp/carbons/xso.py
```diff
@@ -0,0 +1,103 @@
...
+from aioxmpp.utils import namespaces
+
+from ..misc import Forwarded
+from ..stanza import Message
+
+
+namespaces.xep0280_carbons_2 = "urn:xmpp:carbons:2"
+
+
+class Enable(xso.XSO):
```
sebastianriese: Enable and Disable should be registered as IQ payload. Even if they are never received as IQ payloads, they should be registered for documentation purposes. (Perhaps we want a decorator that registers for documentation and future type checking purposes only, and does not enable receiving those stanzas).

horazont: Good catch. Especially since #61 is on our (my?) TODO.

PR #67, opened 2017-04-02
Fix SM counter overflow behaviour

sebastianriese:
* SM counters now wrap modulo 2**32 as is specified by
  XEP-0198

### DIFF
--- a/aioxmpp/stream.py
+++ b/aioxmpp/stream.py
@@ -1035,6 +1035,7 @@ class StanzaStream:
         # now handle stanzas, these always increment the SM counter
         if self._sm_enabled:
             self._sm_inbound_ctr += 1
+            self._sm_inbound_ctr &= 0xffffffff
 
         # check if the stanza has errors
         if exc is not None:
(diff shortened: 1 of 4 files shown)


horazont:
Also, please add a changelog entry.

sebastianriese:
I noticed that I do not understand a detail in the sm_resume tests: They pass as they currently are, but
I would argue, that they are faulty ([code](https://github.com/horazont/aioxmpp/blob/db8784143c84546d30a822691d54f6ce32ba661a/tests/test_stream.py#L3779)):

* the last stanza the simulated remote has acked has sequence number 1. Yet on resumption, the `<resumed>` element carries a sequence number 2.

This could certainly be a me misunderstanding the XEP, and not a bug in the implementation or the tests. The XEP states:
> If the server recognizes the 'previd' as an earlier session that has timed out the server MAY also include a 'h' attribute indicating the number of stanzas received before the timeout.

Our tests, however, fail if the sequence number is set to `1` (which is the number of stanzas acknowledged so far).

horazont:
It may not be sufficient, but I think it is not faulty. The test covers the following scenario: 

* Network between client and server has considerable latency.
* Server acks each stanza immediately with a gratuitous ACK (covered by "After enabling stream management, the client or server can send ack elements at any time over the stream.")
* Client sends three stanzas
* Server receives two stanzas and sends corresponding ACKs
* Client receives the first ACK
* Connection is interrupted and stream resumption takes place

The test as it stands is covered by XEP-0198. If we have issues where the resumption has a counter equal to the last received counter, that definitely needs additional testing and fixes, however, the test as it stands is valid I think.

sebastianriese:
Okay, then the PR should be okay and can be merged. (And the additional test should be added in another PR).

Issue #37, opened 2017-02-07
Reconsider how handlers for message stanzas are registered


horazont:
Currently, it is impossible to register a handler only for a bare JID. In many cases, this is okay, however, the clients own bare JID is a special case here. It represents the account of the client, not any other client like its resources do.

Possible solutions (just brainstorming here, comments welcome):

1. Add a special value referring to the clients own bare JID at any point in time, like what ``None`` currently does implicitly for IQs.

2. Make registration of wildcard handlers explicit and thus allow to register handlers for bare JIDs.

3. …?

horazont:
Has been implemented with the dispatcher classes (see aioxmpp.dispatcher) and the deprecation of ``register_{message,presence}_callback`` on ``StanzaStream``.

PR #70, opened 2017-04-02
Unified Conversation Interface, Stage 1

horazont:
This is part one of a series of PRs to implement #72.

This PR will tackle the following things:

* [x] Draft the interface; the draft can be seen in the documentation in ``aioxmpp/im/conversation.py``.
* [x] Re-work message distribution (see also #37)
* [x] Implementation for one-on-one chat
* [x] First-class support for Message Carbons for one-on-one chat (see also #42)
* [x] Implementation for XEP-0045 Multi-User-Chats (MUCs)
* [x] First-class support for Message Carbons for MUCs
* [x] Finishing touches on the interface
* [x] ~~Do not break the simpler interfaces existing for sending and receiving messages, specifically including the existing MUC interface (not sure if we can and want to actually do that).~~
* [x] Fix & update MUCClient documentation
* [x] Changelog entry
* [x] Test & fix the MUC-related examples
* [x] Split off changes which do not need to be merged yet (e.g. archives interfaces etc.)
* [x] Basic message tracking (by stanza ID), proper filtering of own messages from received message stream.
* [x] Make sure that MUC PM conversations are created for the full JID, not the bare JID.

This may seem too much; I have considered cutting off after implementing one-on-one chats, however, putting an actual multi-user chat solution into the mix (pun intended) will help with feeling whether the interface is in fact appropriate, thus avoiding large changes down the road.

~~This is still a work-in-progress. While reviews are appreciated, please consider that the whole API may be re-worked in the process. This is a very complex feature and a lot of thought needs to go into the APIs. I also suspect that there are several local maxima for the quality of the API and we might transition from one local maximum to another, breaking everything. So please don’t be disappointed if your review was in the end useless because we decided to re-do the part you reviewed for other reasons.~~

~~And holy cow, **don’t merge this yet**.~~


aioxmpp/im/archive.py
```diff
@@ -0,0 +1,78 @@
...
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program.  If not, see
+# <http://www.gnu.org/licenses/>.
+#
+########################################################################
+import asyncio
+import collections
+
+
+class Pagination:
```
sebastianriese: The interface of `Pagination` should have methods to set the initial position ("get_last", "get_first") to allow traversal in both directions.

aioxmpp/im/p2p.py
```diff
@@ -0,0 +1,179 @@
...
+            try:
+                existing = self._conversationmap[peer.bare()]
+            except KeyError:
+                existing = None
+
+        if     ((msg.type_ == aioxmpp.MessageType.CHAT or
+                 msg.type_ == aioxmpp.MessageType.NORMAL) and
+                msg.body):
+            if existing is None:
+                existing = self._make_conversation(peer.bare())
```
sebastianriese: This makes a new conversation, contradicting the documentation, which says, that no new conversation is created automatically.

aioxmpp/tracking.py
```diff
@@ -30,11 +30,53 @@
...
+
+Option (1) has the appeal that users (applications) do not have to worry about
+properly releasing the tracking objects. However, it has the downside that
+applications have to keep the :class:`MessageeTracker` instance around. Remember
+that connecting to callbacks of an object is *not* enough to keep it alive.
+
+Option (2) is somewhat like file objects work: in theory, you have to close
+them explicitly and manually: if you do not, there is no guarantee when the
+file is actually closed. It is thus a somewhat known Python idiom, and also is
+mor explicit. And it doesn’t break callbacks.
```
sebastianriese: Typo: "mor explicit" => "more explicit"

aioxmpp/im/archive.py
horazont: Hah, I couldn’t even remember that this class was in the PR. Actually I may move this to another branch; those are intended to be used with MAM, so they don’t serve a purpose yet.

That interface definitely will need some work in any case.

aioxmpp/im/p2p.py
horazont: Well spotted. That’s a bug in the documentation.

aioxmpp/muc/service.py
```diff
@@ -697,27 +825,147 @@ def _inbound_muc_user_presence(self, stanza):
...
+        can be used as desired for the message.
+
+        .. seealso::
+
+            :meth:`.AbstractConversation.send_message` for the full interface
+            specification.
+        """
+        msg.type_ = aioxmpp.MessageType.GROUPCHAT
+        msg.to = self._mucjid
+        msg.xep0045_muc_user = muc_xso.UserExt()
```
horazont: I’m following suggestion 1 from <https://mail.jabber.org/pipermail/standards/2017-January/032048.html>:

> Current best practice is for MUC implementations to add a tag to all messages (groupchat and PMs) sent from the MUC:
>
>        <x xmlns="http://jabber.org/protocol/muc#user">
>
> Carbons implementations do check for presence of this tag and disable CC'ing accordingly. We need to codify this, and it belongs into both XEP-0045 (adding the <x/> tag) and into XEP-0280 (using the tag as a filter criterion). I'm willing to create PRs for both once consensus is reached ("45 version bump", anyone?).

I also realise just now that this doesn’t affect us, because we do not "send from the MUC" and are thus not the MUC implementation Georg is talking about there. However, it doesn’t hurt and may in fact help if the MUC we’re using doesn’t follow this suggestion and also does not strip the ``<x/>`` we’re adding.

aioxmpp/muc/service.py
```diff
@@ -524,42 +612,72 @@ def _resume(self):
...
+        self.on_muc_resume()
 
-    def _inbound_message(self, stanza):
+    def _match_tracker(self, message):
+        try:
+            tracker = self._tracking_by_id[message.id_]
+        except KeyError:
+            try:
+                tracker = self._tracking_by_sender_body[
+                    message.from_, message.body.get(None)
```
horazont: You’re right, this needs fixing.

aioxmpp/muc/service.py
sebastianriese: Ok, that makes sense. XMPP has some quite arcane corners ;).

aioxmpp/muc/service.py
```diff
@@ -618,18 +633,28 @@ def _match_tracker(self, message):
         try:
             tracker = self._tracking_by_id[message.id_]
         except KeyError:
+            key = message.from_, _extract_one_pair(message.body)
```
sebastianriese: Are we sure we get the same pair as when registering the message? Should we not rather search whether any of the messages matches the one we selected when registering the message for tracking.

aioxmpp/muc/service.py
```diff
@@ -39,6 +39,21 @@
 from . import xso as muc_xso
 
 
+def _extract_one_pair(body):
```
sebastianriese: This is a bit problematic. What if the client on the other side is rude and includes empty messages for some language tags, this would severely degrade the tracking.

aioxmpp/muc/service.py
horazont: Yeah, but we have to remember that this is called for *every* inbound message, even if the MUC does not rewrite IDs (because inbound messages except those which are reflections (usually the minority) fail the check above). I would prefer to stick to O(1) checks here. Especially if the MUC is actively evil and actively breaks the tracking: In that case it could make linear checks arbitrarily expensive. This way, it can only break tracking.

aioxmpp/muc/service.py
horazont: Unless the MUC service actively introduces empty ``<body/>`` for xml:lang values we did not send, I don’t think that matters. Also see the discussion below.

sebastianriese: Yes, selecting the lowest element with respect to some consistent order sounds right.

sebastianriese: And: LanguageTags are ordered (by their encapsulated string value).

aioxmpp/muc/service.py
sebastianriese: Note, the runtime of the version using the order is also in some way O(n·m) (n ... number of the messages, m ... number of language entries per message), but all discussed schemes are only O(N) where (N ... size of received messages).

aioxmpp/muc/service.py
sebastianriese: (Note: One should not use sorted but extract the minimal element, since this is O(m) vs. O(m·log(m)))

aioxmpp/muc/service.py
horazont: Right, sorting is also O(n) (per message).

We should probably simply drop the ``from`` address from the dict and always compare with ``self._this_occupant.conversation_jid``. This helps to defeat races by nickname, saves space *and* gives us an O(1) check to weed out the majority of messages.

Then it doesn’t harm *that* much if we actually do the linear scan, either in the ``_extract_one_pair`` or when searching for a match.

Does that sound right?

(Of course, a malicious MUC could still reflect messages with a whole lot of ``<body/>`` elements, but that will only affect this code for those which are coming from our address, making the attack obvious to the user and the MUC useless.)

sebastianriese: Dropping the from_ is problematic in the case of nick changes (and possibly reconnects).

horazont: Nick changes are actually better with dropping ``from_``. Stanzas are strictly ordered; the nickname we have at the time of reflection may be a different one than we think if another client concurrently changes it. However, we know which nickname we have at the time the stanza is received.

sebastianriese: Ok, I restate this, nick changes are problematic with tracking in all cases. You convinced me dropping `from_` is the way to go, it seems to be the most robust heuristic.

PR #81, opened 2017-04-04
Fix EntityCapsService

horazont:
It was broken since e4388f7!

Should we release v0.8.1 due to this? We really need e2e tests for ecaps by the way, that would’ve been caught by those.

The impact is that outgoing capabilities are not generated, even though we advertise the feature.


horazont:
Hang on, I’m going to add e2etests too, while I’m at it. So that this doesn’t break again.

PR #83, opened 2017-04-05
Add a service supporting PEP

### LINKED ISSUE #44: XEP-0163 (Personal Eventing Protocol) support
[XEP-0163 (Personal Eventing Protocol)](https://xmpp.org/extensions/xep-0163.html)

sebastianriese:
This PR is here for discussion, I do not consider it 100% done, I am open to suggestions about improvement. Especially, the check for PEP support is not done aggressive enough (we do not check for PEP support before publishing `+notify` features).

Once this has reached acceptable state, it should be merged an the avatar service should be ported to use this.

This closes #44.


aioxmpp/pep/service.py
```diff
@@ -0,0 +1,238 @@
...
+
+    .. automethod:: claim_pep_node
+
+    .. automethod:: unclaim_pep_node
+
+    """
+    ORDER_AFTER = [
+        aioxmpp.DiscoClient,
+        aioxmpp.DiscoServer,
+        aioxmpp.PubSubClient
```
horazont: Add trailing comma for less diff when this is changed again.

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,238 @@
...
+            :attr:`aioxmpp.PubSubClient.on_item_publish`
+            for the arguments.
+        :param register_feature: Whether to publish the `node_namespace`
+            as feature.
+        :param notify: Whether to register the ``+notify`` feature to
+            receive notification without explicit subscription.
+
+        :raises RuntimeError: if a handler for `node_namespace` is already
+            set.
+        """
```
horazont: I prefer to have (in docstrings):
```python
"""
One paragraph (ideally one line) summary.

:param foo: …
:raises fnord: if baz
:return: bar

Detailed description
"""
```

The advantage is that the level of detail increases section by section and that the reader has a relatively constant offset from the synopsis to the overview of what the parameters do. This is a shortcoming of the official python docs imo (you always have to search to find the docs for a specific parameter).

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,238 @@
...
+        if notify:
+            self._disco_server.unregister_feature(node_namespace + "+notify")
+        if feature:
+            self._disco_server.unregister_feature(node_namespace)
+
+    @asyncio.coroutine
+    def _check_for_pep(self):
+        # XXX: should this be done when the stream connects
+        # and we use the cached result later on (i.e. disable
+        # the PEP service if the server does not support PEP)
```
horazont: Interesting idea. Issues:

1. The DiscoServer does not purge its registrations when the stream is killed. I also think that we should not change that.
2. Users of this service will be slightly surprised to see their registrations gone when PEP support disappears after a reconnect.

The solution would be to provide a callback users of the PEPService can register to to get notified if PEP support is gone and their registrations are dead. I do not like that API though, and I’m open for better solutions.

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,238 @@
...
+    @service.depsignal(aioxmpp.PubSubClient, "on_item_published")
+    def _handle_pubsub_publish(self, jid, node, item, *, message=None):
+        try:
+            handler, _, _ = self._pep_node_claims[node]
+        except KeyError:
+            return
+
+        # TODO: handle empty payloads due to (mis-)configuration of
+        # the node specially.
+        handler(jid, node, item, message=message)
```
horazont: If we don’t care about synchronicity, you could pass ``defer=True`` to ``depsignal`` and make ``_handle_pubsub_publish`` a coroutine. That would allow to retrieve the payload if it isn’t sent.

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        yield from self._check_for_pep()
+        return (yield from self._pubsub.publish(None, node, data, id_=id_))
+
+    def subscribe(self, jid, node):
+        yield from self._pubsub.subscribe(jid, node)
+
+class RegisteredPEPNode:
+    # XXX: do we want to provide metadata in this class?
+    on_event_received = callbacks.Signal()
```
horazont: I’d at least add a ``notify`` property where notification can be turned on and off. Maybe even enabling/disabling the exporting of the feature should be possible.

I would even go that far that the RegisteredPEPNode should be the object used within the PEPService and all normal interfacing which is done through PEPService for enabling/disabling notify etc. should simply be a forward to the RegisteredPEPNode. It should also be possible to obtain the RegisteredPEPNode when claiming a namespace so that callbacks can be used without the descriptor (useful if you implement a PEP application in the application layer, without services).

Using the descriptor then is just a simple interface to obtain the RegisteredPEPNode.

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,238 @@
...
+
+        :returns: The PubSub id of the published item or
+            :data:`None` if it is unknown.
+
+        """
+        yield from self._check_for_pep()
+        return (yield from self._pubsub.publish(None, node, data, id_=id_))
+
+    def subscribe(self, jid, node):
+        yield from self._pubsub.subscribe(jid, node)
```
horazont: Methods for enabling/disabling notify at some later point would be great.

tests/pep/test_e2e.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @blocking_timed
+    @asyncio.coroutine
+    def test_claim_node_and_get_notification(self):
+        done = asyncio.Event()
+        def handler(jid, node, item, *, message=None):
+            self.assertEqual(jid, self.client.local_jid.bare())
+            self.assertEqual(node, "urn:example:payload")
+            self.assertEqual(item.registered_payload.data, EXAMPLE_TEXT)
+            done.set()
```
horazont: Have you tested that the assertions work? I’m afraid they don’t – this is a Signal callback, and Signal swallows exceptions (and disconnects the handler). This will probably also affect AssertionErrors.

My preferred way to do is is this:
```python
fut = asyncio.Future()
def handler(jid, node, item, *, message=None):
    fut.set_result((jid, node, item))

# […]

jid, node, item = yield from fut

self.assertEqual(jid, self.client.local_jid.bare())
self.assertEqual(node, "urn:example:payload")
self.assertEqual(item.registered_payload.data, EXAMPLE_TEXT)
```

tests/pep/test_e2e.py
```diff
@@ -0,0 +1,122 @@
...
+        p = self.client.summon(aioxmpp.pep.PEPClient)
+        p.claim_pep_node("urn:example:payload", handler, notify=True)
+        # this is necessary, otherwise the +notify feature will not be
+        # sent to the server.
+        yield from presence.resend_presence()
+        payload = ExamplePayload()
+        payload.data = EXAMPLE_TEXT
+        yield from p.publish("urn:example:payload", payload)
+        yield from done.wait()
+        p.unclaim_pep_node("urn:example:payload")
```
horazont: Does this matter? The client is thrown away anyways.

tests/pep/test_e2e.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @blocking_timed
+    @asyncio.coroutine
+    def test_get_notification(self):
+        done = asyncio.Event()
+        def handler(jid, node, item, *, message=None):
+            self.assertEqual(jid, self.client.local_jid.bare())
+            self.assertEqual(node, "urn:example:payload")
+            self.assertEqual(item.registered_payload.data, EXAMPLE_TEXT)
+            done.set()
```
horazont: Same thing as above.

tests/pep/test_service.py
```diff
@@ -0,0 +1,246 @@
...
+                id_="example-id",
+            ))
+
+            self.s._check_for_pep.assert_called_once_with()
+            self.pubsub.publish.assert_called_once_with(
+                None,
+                "urn:example",
+                unittest.mock.sentinel.data,
+                id_="example-id",
+            )
```
horazont: That’s also a way to do it. FWIW, I prefer to have the mocks in a variable; you get them from ``enter_context``:

```python
            publish = stack.enter_context(
                unittest.mock.patch.object(
                    self.pubsub,
                    "publish",
                    CoroutineMock(),
                )
            )
```

Do it however you prefer though; advantage of my way is that you can do it outside of the CM.

tests/pep/test_service.py
```diff
@@ -0,0 +1,246 @@
...
+    def test_claim_pep_node_twice(self):
+        handler1 = unittest.mock.Mock()
+        handler2 = unittest.mock.Mock()
+        with unittest.mock.patch.object(
+                self.disco_server,
+                "register_feature"):
+            self.s.claim_pep_node("urn:example", handler1)
+            with self.assertRaisesRegex(RuntimeError,
+                    "^setting handler for already handled namespace$"):
+                self.s.claim_pep_node("urn:example", handler2)
```
horazont: Maybe add an assertion to ensure that register_feature isn’t called in the second attempt. I always like to explicitly check strong exception safety.

aioxmpp/pep/service.py
sebastianriese: Ah, that is a nice feature.

aioxmpp/pep/service.py
sebastianriese: Yes, I had thought about something like that as well. You get a token for your namespace claim on calling `claim_pep_node`, which is also used later on, when unclaiming the namespace, this would be a perfect match for the object to be accessible in the descriptor.

tests/pep/test_e2e.py
sebastianriese: Uppsi. This also affects the avatar tests, this should be fixed.

tests/pep/test_e2e.py
sebastianriese: Probably it doesn't matter, just wanted to be on the safe side to prevent interference between the test cases.

tests/pep/test_service.py
sebastianriese: Yes, I am not entirely sure why I adopted that style. I thought I had a reason and continued with it. Probably your style is cleaner.

tests/pep/test_e2e.py
horazont: The client is definitely recycled between test cases. The whole client is thrown away and an entirely new one is picked.

However, coming to think about it, we should probably clean it up anyways. Test cases which do not use the anonymous provisioner (I plan to add more provisioners, e.g. one where you put a fixed list of accounts from which it sources clients) could otherwise break.

aioxmpp/pep/service.py
horazont: Note that this really breaks synchronicity. So it makes the event useless e.g. for the Avatar-Race-Condition-Detection-Usecase.

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,363 @@
...
+            self,
+            node_namespace,
+            register_feature=register_feature,
+            notify=notify,
+        )
+
+        finalizer = weakref.finalize(
+            registered_node,
+            weakref.WeakMethod(registered_node._unregister)
+        )
```
horazont: I thought the whole point of finalize is to keep a reference alive? Why do we use a WeakMethod there?

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,363 @@
...
+        # we cannot guarantee that disco is not cleared up already
+        finalizer.atexit = False
```
horazont: We could. Keep the finalizer around and run it from ``_shutdown`` (see Service docs). That’s good enough.

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,363 @@
...
+        return self._notify
+
+    @notify.setter
+    def notify(self, value):
+        if self._closed:
+            raise RuntimeError(
+                "modifying a closed RegisteredPEPNode is forbidden"
+            )
+        # XXX: do we want to do strict type checking here?
+        if (not value) == self._notify:
```
horazont:
```python
if bool(value) == bool(self._notify):
    return
```

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,363 @@
...
+        return self._feature_registered
+
+    @feature_registered.setter
+    def feature_registered(self, value):
+        if self._closed:
+            raise RuntimeError(
+                "modifying a closed RegisteredPEPNode is forbidden"
+            )
+        # XXX: do we want to do strict type checking here?
+        if (not value) == self._feature_registered:
```
horazont: Same as above.

aioxmpp/pep/service.py
sebastianriese: No, the point of finalize is to register a finalizer (which is better behaved than `__del__`). The first argument is weakrefed, the finalizer is not called until only weakrefs remain. If we have a finalizer method, the reference from that method will keep the node alive (the ref to the callable is not weak), thereby delaying collection indefinitely!

aioxmpp/pep/service.py
sebastianriese: Okay, that is a good idea. We will however leak finalizers of the already collected claims.

aioxmpp/pep/service.py
sebastianriese: self._notify is guaranteed to be a bool, but admittedly using bool is safer.

aioxmpp/pep/service.py
sebastianriese: This issue is still open and I have not found a suitable solution. I am afraid this is quite tricky.

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,363 @@
...
+            :data:`None` if it is unknown.
+
+        If no `id_` is given it is generated by the server (and may be
+        returned).
+        """
+        yield from self._check_for_pep()
+        return (yield from self._pubsub.publish(None, node, data, id_=id_))
+
+
+class RegisteredPEPNode:
```
sebastianriese: I am not totally confident as to the robustness of the unregistration process in under all circumstances, the interdependence of the code parts feels to be to complex.

horazont: Is this something which can be helped with more tests? If not, why not? If so, how complex would those tests be? I will take a closer look in any case.

aioxmpp/pep/service.py
```diff
@@ -0,0 +1,363 @@
...
+        registered_node = RegisteredPEPNode(
+            self,
+            node_namespace,
+            register_feature=register_feature,
+            notify=notify,
+        )
+
+        finalizer = weakref.finalize(
+            registered_node,
+            weakref.WeakMethod(registered_node._unregister)
```
horazont: Shouldn’t this call ``close`` instead of ``_unregister``? Two benefits: If it has been closed, it won’t fail additionally and it will clean up the internal stuff in PEPService too.

aioxmpp/pep/service.py
sebastianriese: The tests are relatively extensive already and should get most cases. I feel uncomfortable about the complexity anyway (interaction of the close method and the weakref cleanup). Especially, because disco_server.{un,}register_feature raise exceptions if they are called twice.

aioxmpp/pep/service.py
sebastianriese: No it should definitely not be close: close removes the namespace explicitely from the claimed_pep_nodes dict, this leads to exceptions in combination with the automatic removal of the nodes due to the WeakValueDicationary.

horazont: Ahh, you’re still using the WeakValueDictionary. I missed that; I thought the idea was to put weakrefs in the dict ourselves, make the checks on the liveness before using/returning them and use the finalizer to clean them up there. In that case ``close`` would be the right thing. This would also help with avoiding double calls to DiscoServer, wouldn’t it?

sebastianriese: The weakref callback is called with the weakref as argument, when the referent is no longer available. From the docs:
> If callback is provided and not None, and the returned weakref object is still alive, the callback will be called when the object is about to be finalized; the weak reference object will be passed as the only parameter to the callback; the referent will no longer be available.

That's the reason I switched back to `WeakValueDict` and use `weakref.finalize`.

horazont: Ah, I see. Complex stuff. I think in case of doubt we could also live with not having that weakref magic at all. Just advise to either close claims properly or simply use the descriptor. The descriptor is the preferred interface anyways, and it is properly closed during shutdown.

aioxmpp/pep/service.py
horazont: My point here is actually about two things: first, that ``not value`` and second that it’s clearer to if-and-return than if-and-do-things if possible.

I have no strong opinion on bool-ising self._notify.

horazont:
This looks stalled, which changes are you still planning to do? I kind of lost track.

horazont:
What are the exact semantics of ``register_pep_node``? Is it compatible to "split" servies such as Avatar, which is split in AvatarServer and AvatarClient?

sebastianriese:
register_pep_node certainly causes problems with split services (although it should work with the Avatar example). I guess we should extend the interface to allow a "server" and a "client" claim per namespace.

horazont:
I’m not sure that’s worth the trouble. Maybe rather re-consider whether split services make sense in the PEP context.

If the services allow to turn off the client part (by disabling notify), that’s good enough for me. Otherwise, the client part usually has no runtime cost.

sebastianriese:
And the server part of a PEP service usually has no runtime cost since it must be called explicitly. However, the different interface use for sending and retrieving avatar images would be more confusing if both were supplied by the same service. I am not entirely sure, that disallowing split services is the Right Thing.

aioxmpp/pep/service.py
sebastianriese: Okay, I think the correct solution for now is to ignore this issue. Cases where PEP support vanishes during a reconnect should be *very* rare (as it requires a reconfigured server). We might want to add a call to check_for_pep when registering notify features, but even that isn't that important, we will simply not get notifications, and that's it. (And I do not see a clean solution that works in all cases).

PR #94, opened 2017-04-18
Increase timeouts to help preventing spurious failures in Travis

horazont:


horazont:
FWIW, I don’t think this is complete; but it should prevent two spurious failures I had in #93.

horazont:
@sebastianriese I don’t know; ``run_coroutine`` is the go-to place to increase timeouts because it is used ~everywhere and has a default timeout of 1s. The ``run_coroutine``-based test failures may reduce by an order of magnitude I hope (since I doubled the timeout and it isn’t set very tight in the first place).

For other issues, I plan to increase timeouts as they happen.

Issue #11, opened 2016-06-21
Convert code base to Python 3.5


horazont:
Make use of new features in Python 3.5 and fix deprecations.

- [ ] Use ``asyncio.ensure_future`` instead of ``asyncio.async``
- [ ] Use ``async def `` instead of ``@asyncio.coroutine``
- [ ] Implement more features using ``__aenter__`` / ``__aexit__``

horazont:
This issue will be re-bound to another milestone when it is determined when we will start requiring Python 3.5.

horazont:
Re-binding this to v0.11:

* Ubuntu LTS (18.04) has Python 3.6
* Debian Stable (stretch) has Python 3.5

PR #98, opened 2017-05-01
Support for XEP-0390 (Entity Capabilities 2.0)

horazont:
This depends on #97!

This implements support for emitting and processing XEP-0390 capabilities, as well as the possibility to turn emission and processing for XEP-0115 and XEP-0390 on and off respectively (they both default to "on").

The following features are still missing and will be implemented in an upcoming PR:

* ~~Keeping the last 3 capability hash sets available for querying (including those from XEP-0115)~~
* Inbound and outbound rate limits

### DIFF
--- a/aioxmpp/entitycaps/service.py
+++ b/aioxmpp/entitycaps/service.py
@@ -33,6 +33,8 @@ import aioxmpp.service
 import aioxmpp.xml
 import aioxmpp.xso
 
+from aioxmpp.utils import namespaces
+
 from . import xso as my_xso
 from . import caps115
 
@@ -242,6 +244,8 @@ class EntityCapsService(aioxmpp.service.Service):
 
     on_ver_changed = aioxmpp.callbacks.Signal()
 
+    xep115_support = disco.register_feature(namespaces.xep0115_caps)
+
     def __init__(self, node, **kwargs):
         super().__init__(node, **kwargs)
 
@@ -250,9 +254,6 @@ class EntityCapsService(aioxmpp.service.Service):
 
         self.disco_server = self.dependencies[disco.DiscoServer]
         self.disco_client = self.dependencies[disco.DiscoClient]
-        self.disco_server.register_feature(
-            "http://jabber.org/protocol/caps"
-        )
 
         self.__115 = caps115.Implementation(self.NODE)
 
@@ -286,9 +287,6 @@ class EntityCapsService(aioxmpp.service.Service):
 
     @asyncio.coroutine
     def _shutdown(self):
-        self.disco_server.unregister_feature(
-            "http://jabber.org/protocol/caps"
-        )
         if self.__current_keys:
             for key in self.__current_keys:
                 self.disco_server.unmount_node(key.node)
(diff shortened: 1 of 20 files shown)


aioxmpp/entitycaps/service.py
```diff
@@ -219,18 +221,42 @@ class EntityCapsService(aioxmpp.service.Service):
...
+        Capability Hash Set is delayed.
+
+        This is used to rate-limit the generation of caps-related presences. It
+        can be set to 0 ot effectively remove that rate-limit if rate-limiting
+        is applied elsewhere, but using a rate-limit is encouraged.
+
+        The Capability Hash Set is updated `update_delay` seconds after the
+        most recent change to the features/identities set on the
+        :class:`aioxmpp.DiscoServer`. That means if you those are changed in
+        intervals of ``0.9*update_delay``, the Capability Hash Set is never
```
horazont: Agreed, I’ll get rid of that rate limiting for other reasons, too. It will be the job of the consumer of on_ver_changed to ensure that proper outbound rate-limiting is applied.

aioxmpp/entitycaps/service.py
```diff
@@ -334,50 +428,86 @@ def lookup_info(self, jid, keys):
...
+        keys = []
+
+        if self.xep390_support:
+            keys.extend(self.__390.extract_keys(presence))
+
+        if self.xep115_support:
+            keys.extend(self.__115.extract_keys(presence))
+
+        if keys:
+            lookup_task = asyncio.async(
```
horazont: Is it?
```python
>>> print(sys.version)
3.4.2 (default, Oct  8 2014, 10:45:20) 
[GCC 4.9.1]
>>> import asyncio
>>> asyncio.ensure_future
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'ensure_future'
```

aioxmpp/hashes.py
```diff
@@ -0,0 +1,263 @@
...
+            "hash algorithm {!r} unknown".format(algo)
+        ) from None
+
+    if not enabled:
+        raise ValueError(
+            "support of {} in XMPP is forbidden".format(algo)
+        )
+
+    try:
+        fun = getattr(hashlib, fun_name)
```
horazont: Hm, I have openssl 1.1 and Python 3.5. hashlib.new("sha3_256") does not work (and hashlib.sha3_256 does not exist).

I also don’t want to try here because is_algo_supported must be reliable and I want it to work without having to create hash objects.

aioxmpp/entitycaps/service.py
sebastianriese: Meh, I should learn to read: It's deprecated in Python 3.4.4. I think we should monkey patch ensure_future for < 3.4.4 (and use getattr and friends, when async becomes a keyword foo.async = bar does not work anymore).

aioxmpp/hashes.py
sebastianriese: (Especially, since hashlib.new is documented as being slow). I think your argumentation is correct, and we should not use hashlib.new.

aioxmpp/entitycaps/service.py
horazont: I don’t think that monkey patching is a good idea here. I was comfortable with monkey patching assert_not_called into Mock, but only because I could do that limited to the test suite.

Monkey patching in a dependency of ours is not something I feel we should be doing. (FWIW, we already do that in MLXC)

sebastianriese: Well, I agree it is uncomfortable, but this is the standard library we modify and only a specific set of versions will be affected (since all later versions have `ensure_future`), on the other hand I have to admit, that we'll discontinue Python 3.4 support once Python versions allowing `async def` will be widespread, which hopefully will be long before a Python version reserving `async` as a keyword will be widespread.

Another workaround could be to use a `aioxmpp.compat` module (and define `ensure_future` there, depending on what is available, Plus, we will probably need a compatibility layer at some point anyways).

horazont: > […] but this is the standard library we modify […]

This is exactly why I refrain from it.

> only a specific set of versions will be affected (since all later versions have ``ensure_future`), on the other hand I have to admit, that we'll discontinue Python 3.4 support once Python versions allowing ``async def`` will be widespread, which hopefully will be long before a Python version reserving ``async`` as a keyword will be widespread.

We have a lot places throughout the code where ``asyncio.async`` is used. This is a matter for #11. As long as we don’t support a version of python which makes ``async`` a keyword globally, I don’t think we should rush to a change here. Once upstream makes the move, we’ll reconsider. In the end, it’ll approximately be a simple ``sed`` over the tree to fix things.

> Another workaround could be to use a aioxmpp.compat module (and define ensure_future there, depending on what is available, Plus, we will probably need a compatibility layer at some point anyways).

I would like to avoid that as far as sensible.

PR #99, opened 2017-05-03
Implement XMPP Ping

### LINKED ISSUE #77: XEP-0199 (XMPP Ping) support
We currently use XEP-0199 pings (to the server) automatically in the Stanza Stream (if Stream Management is not available). However, there’s no proper service which implements them and also we do not reply properly to pings.

Tricky things to note:

* What is the intent of the ping?
* If you only want to know that the entity is there, it isn’t necessary that the entity returns a valid pong; it’s sufficient to see that it has returned the IQ at all.

Discussion in the XSF MUC (logs are currently not available online) seems to have brought the following condition for a valid Pong:

```python
iq.type_ == IQType.RESULT or (
    iq.error.type_ == ErrorType.CANCEL and (
        # response to unhandled IQs as per RFC 6120
        iq.error.condition == (namespaces.stanzas, "service-unavailable") or
        # another response (incorrectly) returned by some implementations
        iq.error.condition == (namespaces.stanzas, "feature-not-implemted")
    )
)
```

We cannot treat every error as valid pong because e.g. a 'wait' typed 'service-unavailable' is used e.g. by prosody if a component required for the ping is not available, in which case it is not sure that the target entity was reached at all.

horazont:
Fixes #77.


aioxmpp/forms/xso.py
```diff
@@ -22,7 +22,7 @@
 import collections
 import enum
 
-import aioxmpp
+import aioxmpp.stanza
 import aioxmpp.xso as xso
```
sebastianriese: Why is this diff in this PR? It is unrelated to the changes in ping handling.

aioxmpp/ping/service.py
```diff
@@ -0,0 +1,84 @@
...
+
+        :param peer: The peer to ping.
+        :type peer: :class:`aioxmpp.JID`
+        :return: error condition
+
+        Send a :xep:`199` ping IQ to `peer` and wait for the reply.
+
+        If the reply is succesful, :data:`None` is returned. Otherwise, the
+        :class:`~aioxmpp.stanza.Error` object of the respnose is returned.
+
```
sebastianriese: Is this the ideal API? Returning `None` or an error stanza will make the use of this method cumbersome.

tests/ping/test_e2e.py
```diff
@@ -0,0 +1,121 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program.  If not, see
+# <http://www.gnu.org/licenses/>.
+#
+########################################################################
+import asyncio
+import uuid
```
sebastianriese: This import is not used.

aioxmpp/forms/xso.py
sebastianriese: (Note: This is about all the changes in this file).

horazont: It is not unrelated. It is there to break an import loop. I don’t know the exact loop off the top of my head, but if you change it back you’ll see the breakage.

aioxmpp/ping/service.py
horazont: I’m not happy about it yet either. Simply letting the exception raise may be better.

aioxmpp/forms/xso.py
sebastianriese: Ok. That makes sense.

aioxmpp/ping/service.py
sebastianriese: Yes, letting the exception propagate seems more natural.

horazont:
Pushed changes, rebased on develop.

Using a more sane way to break the import loop now and letting the exception re-raise instead of returning the error condition.

PR #125, opened 2017-07-03
Implement a highlevel interface to bookmarks

### LINKED ISSUE #123: [XEP-0048/bookmarks] Provide more convenient API
I imagine the following interface:

```python
# emitted when a bookmark is added, either by remote or local party
on_bookmark_added = Signal()
# emitted when a bookmark is removed, either by remote or local party
on_bookmark_removed = Signal()
# emitted when a bookmark is changed, either by remote or local party
on_bookmark_changed = Signal()

async def add_bookmark(self, bookmark_xso):
    """
    Add or update a bookmark to server-side storage.
    """

async def update_bookmark(self, old, new):
    """
    Replace a bookmark.

    This operation fails if no bookmark equal to ``old`` is found. 
    Otherwise, ``old`` is replaced by ``new``. ``old`` must be equal 
    to ``new``.
    """

async def remove_bookmark(self, bookmark_xso):
    """
    Remove a bookmark from server-side storage.
    """

async def get_bookmarks(self):
    """
    Return all bookmarks from server-side storage.
    """

async def sync(self):
    """
    Force fetch of bookmarks from the server, unless PubSub is used. Emits events.
    """
```

Notes:
* the events shall not fire spuriously when multiple writes are required due to race conditions
* if pubsub storage is used, ``sync`` should be a no-op
* if private XML storage is used, ``sync`` must be called periodically to be able to emit the events
* add, remove and get also cause events to be emitted, since all will require reads of the remote storage
* equivalence of bookmarks depends on the type:

   * for ``<conference/>``: use bare jid for equality
   * for ``<url/>``: use url for equality

* equivalent bookmarks are not duplicated on ``add``, only updated
* equivalence is used to determine which bookmark to remove
* the user should not have to take care of performing multiple writes if a race condition with another entity occurs.
* a ``sync`` shall happen implicitly when the stream is established

sebastianriese:
Fixes #123.

### DIFF
--- a/aioxmpp/bookmarks/__init__.py
+++ b/aioxmpp/bookmarks/__init__.py
@@ -71,6 +71,6 @@ might cause data loss due to race conditions.
 
 """
 
-from .xso import (Storage, Bookmark, Conference, URL,
-                  as_bookmark_class)  # NOQA
+from .xso import (Storage, Bookmark, Conference, URL,  # NOQA
+                  as_bookmark_class)
 from .service import BookmarkClient  # NOQA
--- a/aioxmpp/bookmarks/__init__.py
+++ b/aioxmpp/bookmarks/__init__.py
@@ -48,9 +48,11 @@ The following XSOs are used to represent an manipulate bookmark lists.
 
 .. autoclass:: URL
 
-The following is used internally as the XSO container for bookmarks. If
-you want to support custom bookmark types, they have to be registered
-as children with the :attr:`Storage.bookmarks` descriptor:
+To register custom bookmark classes use:
+
+.. autofunction:: as_bookmark_class
+
+The following is used internally as the XSO container for bookmarks.
 
 .. autoclass:: Storage
 
@@ -69,5 +71,6 @@ might cause data loss due to race conditions.
 
 """
 
-from .xso import (Storage, Bookmark, Conference, URL)  # NOQA
+from .xso import (Storage, Bookmark, Conference, URL,
+                  as_bookmark_class)  # NOQA
 from .service import BookmarkClient  # NOQA
--- a/aioxmpp/bookmarks/__init__.py
+++ b/aioxmpp/bookmarks/__init__.py
@@ -64,7 +64,7 @@ Notes on usage
 It is highly recommended to interact with the bookmark client via the
 provided signals and the get-modify-set methods
 :meth:`~BookmarkClient.add_bookmark`,
-:meth:`~BookmarkClient.remove_bookmark` and
+:meth:`~BookmarkClient.discard_bookmark` and
 :meth:`~BookmarkClient.update_bookmark`.  Using
 :meth:`~BookmarkClient.set_bookmarks` directly is error prone and
 might cause data loss due to race conditions.
--- a/aioxmpp/bookmarks/__init__.py
+++ b/aioxmpp/bookmarks/__init__.py
@@ -19,7 +19,8 @@
 # <http://www.gnu.org/licenses/>.
 #
 ########################################################################
-""":mod:`~aioxmpp.bookmarks` – Bookmark support (:xep:`0048`)
+"""
+:mod:`~aioxmpp.bookmarks` – Bookmark support (:xep:`0048`)
 ##########################################################
 
 This module provides support for storing and retrieving bookmarks on
@@ -37,50 +38,36 @@ Service
 XSOs
 ====
 
-The following XSOs are used to represent an manipulate bookmark lists.
+All bookmark types must adhere to the following ABC:
 
-.. autoclass:: Storage
+.. autoclass:: Bookmark
+
+The following XSOs are used to represent an manipulate bookmark lists.
 
 .. autoclass:: Conference
 
 .. autoclass:: URL
 
+The following is used internally as the XSO container for bookmarks. If
+you want to support custom bookmark types, they have to be registered
+as children with the :attr:`Storage.bookmarks` descriptor:
+
+.. autoclass:: Storage
+
 Notes on usage
 ==============
 
-It is important to use this class carefully to prevent race conditions
-with other clients modifying the bookmark data (unfortunately, this is
-not entirely preventable due to the design of the bookmark protocol).
+.. currentmodule:: aioxmpp
 
-The recommended practice is to modify the bookmarks in a get, modify,
-set fashion, where the modify step should be as short as possible
-(that is, should not depend on user input).
+It is highly recommended to interact with the bookmark client via the
+provided signals and the get-modify-set methods
+:meth:`~BookmarkClient.add_bookmark`,
+:meth:`~BookmarkClient.remove_bookmark` and
+:meth:`~BookmarkClient.update_bookmark`.  Using
+:meth:`~BookmarkClient.set_bookmarks` directly is error prone and
+might cause data loss due to race conditions.
 
-The individual bookmark classes support comparison by value. This
-allows useful manipulation of the bookmark list in agreement with
-the get-modify-set pattern. Where we reference the objects to operate
-on by with the value retrieved in an earlier get.
-
-Removal::
-
-  storage = bookmark_client.get_bookmarks()
-  storage.bookmarks.remove(old_bookmark)
-  bookmark_client.set_bookmarks(storage)
-
-Modifying::
-
-  storage = bookmark_client.get_bookmarks()
-  i = storage.bookmarks.index(old_bookmark)
-  storage.bookmarks[i].name = "New Shiny Name"
-  storage.bookmarks[i].nick = "new_nick"
-  bookmark_client.set_bookmarks(storage)
-
-Adding::
-
-  storage = bookmark_client.get_bookmarks()
-  storage.bookmarks.append(new_bookmark)
-  bookmark_client.set_bookmarks(storage)
 """
 
-from .xso import (Storage, Conference, URL)  # NOQA
+from .xso import (Storage, Bookmark, Conference, URL)  # NOQA
 from .service import BookmarkClient  # NOQA
--- a/aioxmpp/bookmarks/service.py
+++ b/aioxmpp/bookmarks/service.py
@@ -21,11 +21,8 @@
 ########################################################################
 import asyncio
 
-import aioxmpp
 import aioxmpp.callbacks as callbacks
 import aioxmpp.service as service
-import aioxmpp.disco as disco
-import aioxmpp.pubsub as pubsub
 import aioxmpp.private_xml as private_xml
 
 from . import xso as bookmark_xso
@@ -33,19 +30,65 @@ from . import xso as bookmark_xso
 
 # TODO: use private storage in pubsub where available.
 # TODO: sync bookmarks between pubsub and private xml storage
+# TODO: do we need merge-capabilities to reconcile the bookmarks
+# from different sources (local bookmark storage, pubsub, private xml
+# storage)
 class BookmarkClient(service.Service):
     """
     Supports retrieval and storage of bookmarks on the server.
     It currently only supports :xep:`Private XML Storage <49>` as
     backend.
 
+    There is the general rule *never* to modify the bookmark instances
+    retrieved from this class (either by :meth:`get_bookmarks` or as
+    an argument to one of the signals). If you need to modify a bookmark
+    for use with :meth:`update_bookmark` use :func:`copy.copy` to create
+    a copy.
+
+    .. automethod:: sync
+
     .. automethod:: get_bookmarks
 
     .. automethod:: set_bookmarks
 
+    The following methods change the bookmark list in a get-modify-set
+    pattern, to mitigate the danger of race conditions and should be
+    used in most circumstances:
+
+    .. automethod:: add_bookmark
+
+    .. automethod:: remove_bookmark
+
+    .. automethod:: update_bookmark
+
+
+    The following signals are provided that allow tracking the changes to
+    the bookmark list:
+
+    .. signal:: on_bookmark_added(added_bookmark)
+
+        Fires when a new bookmark is added.
+
+    .. signal:: on_bookmark_removed(removed_bookmark)
+
+        Fires when a bookmark is removed.
+
+    .. signal:: on_bookmark_changed(old_bookmark, new_bookmark)
+
+        Fires when a bookmark is changed.
+
+    .. note:: A heuristic is used to determine the change of bookmarks
+              and the reported changes may not directly reflect the
+              used methods, but it will always be possible to
+              construct the list of bookmarks from the events. For
+              example, when using :meth:`update_bookmark` to change
+              the JID of a :class:`Conference` bookmark a removed and
+              a added signal will fire.
+
     .. note:: The bookmark protocol is prone to race conditions if
               several clients access it concurrently. Be careful to
-              use a get-modify-set pattern.
+              use a get-modify-set pattern or the provided highlevel
+              interface.
 
     .. note:: Some other clients extend the bookmark format. For now
               those extensions are silently dropped by our XSOs, and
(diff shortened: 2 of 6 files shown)


aioxmpp/bookmarks/__init__.py
```diff
@@ -37,50 +38,36 @@
...
 
-Notes on usage
-==============
-
-It is important to use this class carefully to prevent race conditions
-with other clients modifying the bookmark data (unfortunately, this is
-not entirely preventable due to the design of the bookmark protocol).
+The following is used internally as the XSO container for bookmarks. If
+you want to support custom bookmark types, they have to be registered
+as children with the :attr:`Storage.bookmarks` descriptor:
```
horazont: Can we have a simple example on how that works?

aioxmpp/bookmarks/__init__.py
```diff
@@ -37,50 +38,36 @@
...
-Modifying::
+Notes on usage
+==============
 
-  storage = bookmark_client.get_bookmarks()
-  i = storage.bookmarks.index(old_bookmark)
-  storage.bookmarks[i].name = "New Shiny Name"
-  storage.bookmarks[i].nick = "new_nick"
-  bookmark_client.set_bookmarks(storage)
+.. currentmodule:: aioxmpp
```
horazont: Why that?

aioxmpp/bookmarks/service.py
```diff
@@ -21,31 +21,66 @@
...
 class BookmarkClient(service.Service):
     """
     Supports retrieval and storage of bookmarks on the server.
     It currently only supports :xep:`Private XML Storage <49>` as
     backend.
 
+    There is the general rule *never* to modify the bookmark instances
+    retrieved from this class (either by :meth:`get_bookmarks` or as
+    an argument to one of the signals). If you need to modify a bookmark
+    for use with :meth:`update_bookmark` use :func:`~copy.copy` to create
```
horazont: For references to non-aioxmpp functions and modules, I generally prefer to have the fully-qualified name listed. That makes things immediately clear and stops the reader from having to interrupt flow to figure out if that is an aioxmpp-specific ``copy`` or the python standard library ``copy``.

aioxmpp/bookmarks/service.py
```diff
@@ -57,29 +92,256 @@ class BookmarkClient(service.Service):
...
         Set the bookmarks stored on the server.
         """
-        if not isinstance(bookmarks, bookmark_xso.Storage):
-            raise TypeError(
-                "set_bookmarks only accepts bookmark.Storage objects"
+        storage = bookmark_xso.Storage()
+        storage.bookmarks[:] = bookmarks
+        yield from self._private_xml.set_private_xml(storage)
+
+    def _diff_emit_update(self, new_bookmarks):
```
horazont: I trust your unittests on this one.

aioxmpp/bookmarks/service.py
```diff
@@ -57,29 +92,256 @@ class BookmarkClient(service.Service):
...
+        return bookmarks
+
+    @asyncio.coroutine
+    def set_bookmarks(self, bookmarks):
+        """
+        Store the sequence of bookmarks `bookmarks`.
+
+        Causes signals to be fired to reflect the changes.
+
+        .. note:: This should normally not be used unless you have to
```
horazont: Can we rephrase that so that the emphasis is on the fact that this is only safe when no other clients are modifying the bookmarks? In all other cases, the normal methods should really be used.

aioxmpp/bookmarks/service.py
```diff
@@ -57,29 +92,256 @@ class BookmarkClient(service.Service):
...
+
+        Already existant bookmarks are not added twice.
+        """
+        bookmarks = yield from self._get_bookmarks()
+        for bookmark in bookmarks:
+            if bookmark == new_bookmark:
+                break
+        else:
+            bookmarks.append(new_bookmark)
+        yield from self._set_bookmarks(bookmarks)
```
horazont: This is not complete yet, I think. We should be checking that the remote bookmarks actually contain the bookmark we wanted to add, and if not, add it again (using the freshly retrieved set of bookmarks as a baseline). With a configurable limit of re-tries (3 maybe?). (We will need this re-tried write for pubsub, too.) Only after all that is done, the _diff_emit_update should be called.

If you don’t want to put this in this PR, I’m fine with merging this and handling it in a future update (but it is part of #123, so this PR would not close #123).

aioxmpp/bookmarks/service.py
```diff
@@ -57,29 +92,256 @@ class BookmarkClient(service.Service):
...
+        bookmarks = yield from self._get_bookmarks()
+        result = []
+        not_removed = True
+        for bookmark in bookmarks:
+            if not_removed and bookmark == bookmark_to_remove:
+                not_removed = False
+                continue
+            else:
+                result.append(bookmark)
+        yield from self._set_bookmarks(result)
```
horazont: The same holds as for ``add_bookmark``.

aioxmpp/bookmarks/service.py
```diff
@@ -57,29 +92,256 @@ class BookmarkClient(service.Service):
...
+        `new` is added if no bookmark matching `old` exists.
+        """
+        bookmarks = yield from self._get_bookmarks()
+        for i, bookmark in enumerate(bookmarks):
+            if bookmark == old:
+                bookmarks[i] = new
+                break
+        else:
+            bookmarks.append(new)
+        yield from self._set_bookmarks(bookmarks)
```
horazont: The same holds as for ``add_bookmark``.

aioxmpp/bookmarks/xso.py
```diff
@@ -30,7 +31,96 @@
...
+
+    .. autoproperty:: secondary
+
+    .. autoproperty:: name
+
+    Equality is defined in terms of those properties:
+
+    .. automethod:: __eq__
+
+    It is highly recommenden not to redefine :meth:`__eq__` in a
```
horazont: Typo (recommenden sounds oddly dutch).

aioxmpp/bookmarks/xso.py
```diff
@@ -30,7 +31,96 @@
...
+    subclass, if you do so make sure that the following law
+    holds for :meth:`__eq__`, :attr:`primary` and :attr:`secondary`:
+
+        (type(a) == type(b) and
+         a.primary == b.primary and
+         a.secondary == b.secondary)
```
horazont: Is the formatting here intended? It doesn’t look that nice when rendered in HTML, so maybe use ``::`` to make it code formatted:

>     holds for :meth:`__eq__`, :attr:`primary` and :attr:`secondary`::
>         
>         (type(a)…
>
>     if and only if::

aioxmpp/bookmarks/xso.py
```diff
@@ -30,7 +31,96 @@
...
+        and for the get-modify-set methods.
+        """
+        return (type(self) == type(other) and
+                self.primary == other.primary and
+                self.secondary == other.secondary)
+
+    @abstractproperty
+    def primary(self):
+        """
+        Return the primary category of the bookmark.
```
horazont: Would anything break if an application returned a tuple at this property?

docs/api/changelog.rst
```diff
@@ -15,6 +15,12 @@ Version 0.10
   growth). This split serves the maintainability of the code and offers
   opportunities for better error detection.
 
+* **Breaking change**: :meth:`aioxmpp.BookmarkService.get_bookmarks`
+  now returns a list instead of a :class:`aioxmpp.bookmarks.Storage`
+  and :meth:`aioxmpp.BookmarkService.set_bookmarks` now accepts a
+  list. The list returned by the get method and its contents *must
```
horazont: I suggest: *The list […] and its elements must not be modified*

aioxmpp/bookmarks/xso.py
sebastianriese: Nope, that was not intended, forgot the `::`.

aioxmpp/bookmarks/xso.py
sebastianriese: No. I should document, that this is explicitly allowed (and that the inner structure of the primary category is opaque).

aioxmpp/bookmarks/__init__.py
sebastianriese: BookmarkClient does not resolve via aioxmpp.bookmarks (I don't get why, but compare above, where `.. currentmodule:: aioxmpp` is used to document BookmarkClient).

horazont: Ugh, right. You’re doing the right thing. Nothing to see here, move on.

sebastianriese:
Tests for the retry behaviour are still missing (lowering the coverage considerable), everything else is up and running! All the old tests pass without modification.

aioxmpp/bookmarks/service.py
```diff
@@ -290,58 +303,107 @@ def sync(self):
...
+                              after `max_retries` retries.
         """
-        bookmarks = yield from self._get_bookmarks()
-        for bookmark in bookmarks:
-            if bookmark == new_bookmark:
-                break
-        else:
-            bookmarks.append(new_bookmark)
-        yield from self._set_bookmarks(bookmarks)
+        with (yield from self._lock):
```
horazont: I think we need a ``try/finally`` with ``_diff_emit_update`` around the whole thing. Otherwise, we get inconsistent state in a listener if one of the getters/setters raises for any reason.

aioxmpp/bookmarks/service.py
```diff
@@ -290,58 +303,107 @@ def sync(self):
...
-            else:
-                result.append(bookmark)
-        yield from self._set_bookmarks(result)
-        self._diff_emit_update(result)
+        with (yield from self._lock):
+            bookmarks = yield from self._get_bookmarks()
+            occurences = bookmarks.count(bookmark_to_remove)
+            error = False
+
+            if occurences:
```
horazont: *If* we need the try/finally (see above), we can rewrite this much nicer:
```python
with (…):
    try:
        …
        if not occurences:
            return
        …
    finally:
        self._diff_emit_update(bookmarks)
```

aioxmpp/bookmarks/service.py
```diff
@@ -290,58 +303,107 @@ def sync(self):
...
+
+            self._diff_emit_update(bookmarks)
 
-        self._diff_emit_update(bookmarks)
+            if new_bookmark not in bookmarks:
+                raise RuntimeError("Could not add bookmark")
 
     @asyncio.coroutine
-    def remove_bookmark(self, bookmark_to_remove):
+    def remove_bookmark(self, bookmark_to_remove, max_retries=3):
```
horazont: Same as above.

aioxmpp/bookmarks/service.py
```diff
@@ -290,58 +303,107 @@ def sync(self):
         yield from self.get_bookmarks()
 
     @asyncio.coroutine
-    def add_bookmark(self, new_bookmark):
+    def add_bookmark(self, new_bookmark, max_retries=3):
```
horazont: Please add a description of the ``max_retries`` argument here (with ``:param max_retries: …\n:type max_retries: :class:`int` ``) and also make ``max_retries`` kwarg-only.

aioxmpp/bookmarks/service.py
```diff
@@ -290,58 +303,107 @@ def sync(self):
...
+
+            if occurences:
+                bookmarks.remove(bookmark_to_remove)
+                yield from self._set_bookmarks(bookmarks)
+
+                retries = 0
+                bookmarks = yield from self._get_bookmarks()
+                new_occurences = bookmarks.count(bookmark_to_remove)
+                while retries < max_retries:
+                    if new_occurences < occurences:
```
horazont: Well done, I probably would have deleted two bookmarks in this case! :-)

aioxmpp/bookmarks/service.py
```diff
@@ -290,58 +303,107 @@ def sync(self):
...
+                    error = True
+
+            self._diff_emit_update(bookmarks)
+
+            if error:
+                raise RuntimeError("Could not remove bookmark")
 
     @asyncio.coroutine
-    def update_bookmark(self, old, new):
+    def update_bookmark(self, old, new, max_retries=3):
```
horazont: Same as above.

aioxmpp/bookmarks/service.py
```diff
@@ -290,58 +303,107 @@ def sync(self):
...
-        self._diff_emit_update(bookmarks)
+        def replace_bookmark():
+            for i, bookmark in enumerate(bookmarks):
+                if bookmark == old:
+                    bookmarks[i] = new
+                    break
+            else:
+                bookmarks.append(new)
+
+        with (yield from self._lock):
```
horazont: Again try/finally?

docs/api/changelog.rst
```diff
@@ -46,6 +46,9 @@ Version 0.10
 * Fix (harmless) traceback in logs which could occur when using
   :meth:`aioxmpp.muc.Room.send_message_tracked`.
 
+* You can now register custom bookmark classes with
+  :func:`aioxmpp.bookmarks.as_bookmark_class`.
```
horazont: Add reference to the abstract base class in both the changelog and the function.

aioxmpp/bookmarks/xso.py
```diff
@@ -224,8 +231,28 @@ class Storage(xso.XSO):
...
+
     TAG = (namespaces.xep0048, "storage")
 
     bookmarks = xso.ChildList([URL, Conference])
+
+
+def as_bookmark_class(xso_class):
+    """
+    Decorator to register `xso_class` as a custom bookmark class.
+
```
horazont: Add a ``:raises TypeError: if `xso_class` is not a :class:`Bookmark` subclass``.

horazont:
I think we should have a ``before_stream_established`` or ``on_stream_established`` listener which synchronises the bookmark store.

aioxmpp/bookmarks/service.py
horazont: Also, thinking about the semantics of ``remove_bookmark``, maybe it should be called ``discard_bookmark``? In python, ``remove`` normally raises if an element is not in the collection, ``discard`` doesn’t.

aioxmpp/bookmarks/service.py
sebastianriese: The worst that could happen is outdated state, that will be fixed by the next call to sync. But agreed, it could happen, that we our state does not correspond to the latest received bookmark state.

aioxmpp/bookmarks/service.py
sebastianriese: Yes, I already considered that. But I think that otherwise the semantics are correct (removing *one* bookmark if there are several equal ones and not raising, since there is no reason to complain if we want to remove a bookmark and it's already gone).

horazont: I’m not sure whether you’re arguing for or against ``discard`` here.

aioxmpp/bookmarks/service.py
```diff
@@ -303,107 +303,169 @@ def sync(self):
...
         with (yield from self._lock):
             bookmarks = yield from self._get_bookmarks()
             occurences = bookmarks.count(bookmark_to_remove)
-            error = False
 
-            if occurences:
-                bookmarks.remove(bookmark_to_remove)
-                yield from self._set_bookmarks(bookmarks)
+            try:
+                if occurences:
```
horazont: I think using
```python
if not occurences:
    return
```
is more readable.

aioxmpp/bookmarks/xso.py
```diff
@@ -98,7 +98,7 @@ def primary(self):
         signals. Bookmarks with different type or :attr:`primary`
         keys cannot be identified as changed from/to one another.
         """
-        raise NotImplementedError
+        raise NotImplementedError  # pragma: no cover
```
horazont: TIL!

PR #126, opened 2017-07-04
service: fix is_{dep,attr}signal_handler with defer=True

horazont:
As discovered by @sebastianriese while working on #125.


horazont:
As mode is called by ``connect``, I think we’re on the safe side here. Unless someone makes a callable iterable.

PR #129, opened 2017-07-05
Implement XEP-0085

### LINKED ISSUE #49: XEP-0085 (Chat State Notifications) support
[XEP-0085 (Chat State Notifications)](https://xmpp.org/extensions/xep-0085.html)

sebastianriese:
Fixes #49.

### DIFF
--- a/aioxmpp/chatstates/__init__.py
+++ b/aioxmpp/chatstates/__init__.py
@@ -1,0 +1,58 @@
+########################################################################
+# File name: __init__.py
+# This file is part of: aioxmpp
+#
+# LICENSE
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program.  If not, see
+# <http://www.gnu.org/licenses/>.
+#
+########################################################################
+"""
+:mod:`~aioxmpp.chatstates` – Chat State Notification support (:xep:`0085`)
+##########################################################################
+
+This module provides support to implement :xep:`Chat State
+Notifications <85>`.
+
+XSOs
+====
+
+The module registers an attribute ``xep0085_chatstate`` with`
+:class:`aioxmpp.stanza.Message` to represent the chat state
+notification tags, it takes values from the following enumeration (or
+:data:`None` if no tag is present):
+
+.. autoclass:: ChatState
+
+Helpers
+=======
+
+The module provides the following helper class, that handles the state
+management for chat state notifications:
+
+.. autoclass:: ChatStateManager
+
+Its operation is controlled by one of the chat state strategies:
+
+.. autoclass:: DoNotEmit
+
+.. autoclass:: DiscoverSupport
+
+.. autoclass:: AlwaysEmit
+
+"""
+from .xso import ChatState  # NOQA
+from .utils import (ChatStateManager, DoNotEmit, AlwaysEmit,  # NOQA
+                    DiscoverSupport)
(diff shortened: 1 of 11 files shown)


sebastianriese:
Still TODO: Specify a mechanism to configure the chat state strategy of a conversation (it must be configurable to set the strategy to "DoNotEmit" and MUCs should use "AlwaysEmit" instead of "DiscoverSupport" by default).

horazont:
5. heinäkuuta 2017 18.43.25 UTC+2.00 Sebastian Riese <notifications@github.com> kirjoitti:
>Still TODO: Specify a mechanism to configure the chat state strategy of
>a conversation (it must be configurable to set the strategy to
>"DoNotEmit" and MUCs should use "AlwaysEmit" instead of
>"DiscoverSupport" by default).

I don't thibk that DiscoverSupport is ever a good default due to Message Carbons...

-- 
Sent from my mobile device. Please excuse my brevity.

sebastianriese:
On a side note: I am still not sure, whether the mixin approach is the way to go. The way p2p-conversations are handled here is wrong, we would have to transform Conversation to another mixin, and build up the conversation class that support p2p and chat states by plugging together those two mixins in the correct order. This is especially annoying, since they do not provide the same level of service: on provides actual connectivity, while the other modifies "conversation metadata" building on that. We should seriously consider to rework this to provide a kind of "plugin" mechanism with message filters. This has the advantage, that unused services to not cost memory and processing power, at the cost, that the user will have to configure the wanted features of a conversation.

horazont:
With a plugin mechanism, we have to consider how to handle races when a conversation is started (e.g. by a remote party). There would have to be a very early callback.

I am open for (deeper) modifications of the IM package -- I'm still not convinced that the current system is optimal.
-- 
Sent from my mobile device. Please excuse my brevity.

aioxmpp/im/chatstates.py
```diff
@@ -0,0 +1,93 @@
...
+# <http://www.gnu.org/licenses/>.
+#
+########################################################################
+
+import aioxmpp
+import aioxmpp.im.conversation as conversation
+import aioxmpp.chatstates as chatstates
+
+
+INGOING_STATE_MAP = {
```
horazont: INGOING sounds weird, INBOUND or INCOMING maybe?

aioxmpp/im/chatstates.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def features(self):
+        return (frozenset([conversation.ConversationFeature.SET_STATE]) |
+                super().features)
+
+    def _on_message(self, msg, member, source, tracker=None, **kwargs):
+        if msg.xep0085_chatstate is not None:
+            incoming_state = INGOING_STATE_MAP[msg.xep0085_chatstate]
+            if incoming_state != self.__chatstate_cache.get(member, None):
```
horazont: Is the ``__chatstate_cache`` cleaned when members leave?

aioxmpp/im/chatstates.py
sebastianriese: But it's symmetric with outgoing ;) ...

aioxmpp/im/chatstates.py
sebastianriese: Good point!

horazont:
What are we going to do about this now? Any plans?

horazont:
Okay, I thought about this further, also in the context of MAM.

I think that we should strip the Conversation interface a little, so that it only does the following:

* Provide Member objects representing the members of the conversation
* Handle member management (kick, ban, invite, …), + events
* Associate inbound messages with Members and provide a callback which emits the message plus the associated Member
* Send messages to the conversation, possibly with its own tracking implementation.

Everything beyond that, like interpreting Chat States, Chat Markers, Read Reciepts etc. should be left to the application, with one caveat: Things which implement Message Tracking parts (e.g. Read Reciepts) of course can use the normal tracking API (which is based on stanza filters anyways).

The idea behind this approach is that the same message pipeline which is used to interpret Chat States etc. in a live conversation can be used by the application to interpret these things when querying MAM. It will be up to the application how much and which state is persisted.

While we could provide a very generic implementation in aioxmpp, I think this is territory which should be explored in jclib first.

This also saves us from the "firewall" issue (determine where to send chat states automatically).

For Chat States specifically, I think an object which can be attached to a Conversation (we have the early callback I spoke about earlier now) makes most sense. It would hook into the on_message event to interpret inbound Chat States (without filtering them out) and provide a method to map a Member to a Chat State. It should also have an event signalling a Chat State change and a method to feed stanzas and set chat states manually, e.g. for history replay.

Does that make sense?

sebastianriese:
Now we only supply the necessary XSOs and a helper class for handling ChatStates. The actual work is up to the application.

PR #144, opened 2017-09-21
Synchronous callbacks for IQ handlers

### LINKED ISSUE #143: Provide IQ API with strong ordering guarantees
The current IQ API has a main shortcoming: due to its inherent focus on coroutines, the strong ordering guarantees of the XMPP XML Stream get lost a bit. For example, when registering IQ handlers for two different IQs, we cannot say for sure which IQ handler executes first when the two IQs arrive in the same chunk of buffer handed to the XMLStream protocol. Another example are possible races between the return of the ``send`` coroutine function and execution of IQ request handlers.

This has caused quite a bit of headache in the roster and avatar implementations, since they act on rather large state and that state needs to stay in sync with the servers state to be valid.

To solve this, I propose the following API changes:

1. ``register_iq_request_coro`` -> ``register_iq_request_awaitable`` and allow any function which returns an awaitable to be registered. This includes coroutine functions, but also normal functions returning a Future. The result of the future must adhere to the current specifications for IQ request coroutine function results.

    The ``_coro`` version would become an alias to the ``_awaitable`` version, and is deprecated. To be removed in 1.0.

2. ``send`` gets an additional "callback" argument for use with IQ get/set stanzas. Now I’m not 100% sure. Either:

    1. the callback must be a function returning an awaitable. It is executed right when the IQ result is received. The awaitable is awaited by and the result is returned from ``send``. **or**

    2. the callback must be a normal function without specific return type. It is simply executed right when the IQ result is received. It does not influence the result of ``send``.

  In any case, if the callback is omitted, ``send`` behaves as currently. If the ``callback`` is passed for non-IQ-request stanzas, an error is raised.

This way, we will have a strong order guarantee, since those functions will be called right from the stanza processing loop.

Benefits: we should get around quite a bit of locking, e.g. the global avatar lock (#141) and the roster lock, if we play it well.

horazont:
Essentially what is written in #143:

* Rename ``{un,}register_iq_request{coro => handler}`` and support plain functions returning awaitables as handlers.
* Allow a callback returning an awaitable (or None) when ``send()``-ing IQ requests. The callback is invoked.
* Tack nice warnings and notes on things to explain the reasoning and warn against excessive use of these features.
* Adapt downstream tooling (like the ``iq_handler`` decorator).

Implements/fixes #143.

### DIFF
--- a/aioxmpp/stream.py
+++ b/aioxmpp/stream.py
@@ -911,7 +911,13 @@ class StanzaStream:
                 self.enqueue(response)
                 return
 
-            task = asyncio.async(coro(stanza_obj))
+            try:
+                awaitable = coro(stanza_obj)
+            except Exception as exc:
+                awaitable = asyncio.Future()
+                awaitable.set_exception(exc)
+
+            task = asyncio.async(awaitable)
             task.add_done_callback(
                 functools.partial(
                     self._iq_request_coro_done,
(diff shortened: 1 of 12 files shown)


aioxmpp/stream.py
```diff
@@ -463,6 +457,18 @@ class StanzaStream:
 
     .. automethod:: unregister_presence_callback
 
+    Rarely used registries / deprecated aliases:
+
+    .. automethod:: register_iq_request_coro
+
+    .. automethod:: unregister_iq_request_handler
```
horazont: That’s a typo, it should be ``_coro``. Fixed locally, will push soon.

aioxmpp/stream.py
```diff
@@ -2445,8 +2600,7 @@ def send(self, stanza, *, timeout=None):
                 )
             except asyncio.TimeoutError:
                 raise TimeoutError
-
```
sebastianriese: Not sure why this changed; since the tests pass it probably is correct.

sebastianriese: (This refers to the change on the return argument of send below not shown in the context lines).

horazont: See the ``handler_ok``, it already unwraps the payload from the stanza.

We had to do this because ``cb`` can return anything, not only the stanza.

PR #162, opened 2017-12-12
Make insertion O(1) for aioxmpp.cache.LRUDict

sebastianriese:
* before insertion was O(maxsize*log(maxsize)) which got unacceptably slow
  even for maxsize = 1000
* now we got O(1) retrieval and insertion (getting is now about 1.5 times
  slower!)
* the code is inspired by functools.lru_cache

### DIFF
--- a/aioxmpp/cache.py
+++ b/aioxmpp/cache.py
@@ -33,55 +33,55 @@
 
 import collections.abc
 
-
-PREV = 0
-NEXT = 1
-KEY = 2
-VALUE = 3
+class Link:
+    __slots__ = ("prev", "next_", "key", "value")
 
 
 def _init_linked_list():
-    root = []
-    root[:] = [root, root, None, None]
+    root = Link()
+    root.prev = root
+    root.next_ = root
+    root.key = None
+    root.value = None
     return root
 
 
 def _remove_link(link):
-    link[NEXT][PREV] = link[PREV]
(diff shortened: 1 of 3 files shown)


aioxmpp/cache.py
```diff
@@ -33,6 +33,57 @@
 
 import collections.abc
 
+class Link:
```
horazont: FWIW (this doesn’t need changing), I don’t like the nomenclature. In my textbooks, the parts of a linked list have always been called "Node" or "Element". I find Link confusing, because I would expect it to represent the relationship between two elements, not the element itself.

aioxmpp/cache.py
```diff
@@ -52,25 +103,26 @@ class LRUDict(collections.abc.MutableMapping):
...
+        return (_length(self.__root) == len(self.__links) and
+                _has_consistent_links(self.__root, self.__links))
+
+    def _purge(self):
+        if self.__maxsize is None:
+            return
+
+        while len(self.__links) > self.__maxsize:
+            link = _remove_link(self.__root.prev)
+            del self.__links[link.key]
```
horazont: This is clearly more optimized for the common case that an LRUDict is full. :+1:

aioxmpp/cache.py
sebastianriese: Hmm, probably you are right, node would have been a better name. My association when naming this was a link of a chain.

PR #165, opened 2017-12-13
Implement vCard and vCard Based Avatars

### LINKED ISSUE #138: XEP-0153 (vCard-based Avatars) support
Apparently, there are still clients doing that. The implementation should support reading vCard-based avatars through the AvatarService (obviously without change notifications).

sebastianriese:
This fixes #138.

I am unsure about two points, but I think a review might help to clarify them:
* how to prevent races that result in inconsistent metadata (between calls of get_avatar_metadata
  and the handlers for avatar pushes) – this would be easy if we would have a stanza serial number
  and add this info to the cache as "age" information. Another partial solution would be to bundle 
  requests.
* is the fallback behaviour on vCard avatars done correctly (in which cases do we want to use the 
  fallback). This might take some testing with real implementations to get right (as the responses we
  get may well depend on the XMPP server and client of the peer).

### DIFF
--- a/aioxmpp/avatar/service.py
+++ b/aioxmpp/avatar/service.py
@@ -33,6 +33,7 @@ import aioxmpp.presence as presence
 import aioxmpp.pubsub as pubsub
 import aioxmpp.vcard as vcard
 
+from aioxmpp.cache import LRUDict
 from aioxmpp.utils import namespaces
 
 from . import xso as avatar_xso
@@ -358,11 +359,11 @@ class VCardAvatarDescriptor(AbstractAvatarDescriptor):
         if self._image_bytes is not None:
             return self._image_bytes
 
-        logger.debug("retrieving vCard %s", self._remove_jid)
+        logger.debug("retrieving vCard %s", self._remote_jid)
         vcard = yield from self._vcard.get_vcard(self._remote_jid)
         photo = vcard.get_photo_data()
         if photo is not None:
-            logger.debug("returning vCard avatar %s", self._remove_jid)
+            logger.debug("returning vCard avatar %s", self._remote_jid)
             return photo
 
         raise RuntimeError("Avatar image is not set")
@@ -371,8 +372,7 @@ class VCardAvatarDescriptor(AbstractAvatarDescriptor):
 
 
 class AvatarService(service.Service):
-    """
-    Access and publish User Avatars (:xep:`84`). Fallback to vCard based
+    """Access and publish User Avatars (:xep:`84`). Fallback to vCard based
     avatars (:xep:`153`).
 
     This service provides an interface for accessing the avatar of other
@@ -404,6 +404,21 @@ class AvatarService(service.Service):
 
     .. automethod:: disable_avatar
 
+    Configuration:
+
+    .. autoattribute:: synchronize_vcard
+
+    .. attribute:: avatar_pep
+
+       The PEP descriptor for claiming the avatar metadata namespace.
+       The value is a :class:`~aioxmpp.pep.service.RegisteredPEPNode`,
+       whose :attr:`~aioxmpp.pep.service.RegisteredPEPNode.notify`
+       property can be used to disable or enable the notification
+       feature.
+
+    .. autoattribute:: metadata_cache_size
+       :annotation: = 200
+
     """
 
     ORDER_AFTER = [
@@ -426,7 +441,8 @@ class AvatarService(service.Service):
     def __init__(self, client, **kwargs):
         super().__init__(client, **kwargs)
         self._has_pep_avatar = set()
-        self._metadata_cache = {}
+        self._metadata_cache = LRUDict()
+        self._metadata_cache.maxsize = 200
         self._pubsub = self.dependencies[pubsub.PubSubClient]
         self._pep = self.dependencies[pep.PEPClient]
         self._presence_server = self.dependencies[presence.PresenceServer]
@@ -440,32 +456,53 @@ class AvatarService(service.Service):
         self._publish_lock = asyncio.Lock()
         self._synchronize_vcard = False
         self._vcard_ressource_interference = set()
-        self._vcard_hash = None
+        self._vcard_id = None
+        self._vcard_rehash_task = None
+
+    @property
+    def metadata_cache_size(self):
+        """
+        Maximum number of cache entries in the avatar metadata cache.
+
+        This is mostly a measure to prevent malicious peers from
+        exhausting memory by spamming vCard based avatar metadata for
+        different ressources.
+
+        .. versionadded:: 0.10
+
+        """
+        return self._metadata_cache.maxsize
+
+    @metadata_cache_size.setter
+    def metadata_cache_size(self, value):
+        self._metadata_cache.maxsize = value
 
     @property
     def synchronize_vcard(self):
         """
-        Set this property to true to enable publishing the a vcard avatar.
+        Set this property to true to enable publishing the a vCard avatar.
+
+        This property defaults to false. For the setting true to have
+        effect, you have to publish your avatar with :meth:`publish_avatar_set`
+        or :meth:`disable_avatar` *after* this switch has been set to true.
         """
         return self._synchronize_vcard
 
     @synchronize_vcard.setter
     def synchronize_vcard(self, value):
-        if bool(value) != self._synchronize_vcard:
-            self._vcard_hash = None
         self._synchronize_vcard = bool(value)
 
     @service.depfilter(aioxmpp.stream.StanzaStream,
                        "service_outbound_presence_filter")
     def _attach_vcard_notify_to_presence(self, stanza):
         if not self._vcard_ressource_interference:
-            stanza.xep0153_x = avatar_xso.VCardTempUpdate(self._vcard_hash)
+            stanza.xep0153_x = avatar_xso.VCardTempUpdate(self._vcard_id)
         return stanza
 
     def _handle_notify(self, full_jid, stanza):
         if stanza.xep0153_x is not None:
-            if stanza.photo is not None:
-                if not self._has_pep_avatar[full_jid]:
+            if stanza.xep0153_x.photo is not None:
+                if full_jid not in self._has_pep_avatar:
                     metadata = self._cook_vcard_notify(full_jid, stanza)
                     self.on_metadata_changed(
                         full_jid,
@@ -475,30 +512,42 @@ class AvatarService(service.Service):
 
                 if (full_jid.bare() == self.client.local_jid.bare() and
                     full_jid != self.client.local_jid):
-                    if stanza.photo.lower() != self._vcard_id.lower():
-                        # this may cause a race if there are multiple
-                        # avatar changes in short succession!  As
-                        # first mitigation we may want to cancel
-                        # update_vcard_hash if we set the vcard or get
-                        # a new notification.
+                    if (self._vcard_id is None or
+                        stanza.xep0153_x.photo.lower() !=
+                        self._vcard_id.lower()):
+                        if self._vcard_rehash_task is not None:
+                            self._vcard_rehash_task.cancel()
+
                         self._vcard_id = None
-                        asyncio.get_event_loop().create_task(
-                            self._update_vcard_hash
+                        self._vcard_rehash_task = asyncio.async(
+                            self._calculate_vcard_id()
+                        )
+                        def set_new_vcard_id(fut):
+                            if not fut.cancelled():
+                                self._vcard_id = fut.result()
+                        self._vcard_rehash_task.add_done_callback(
+                            set_new_vcard_id
                         )
 
     @asyncio.coroutine
-    def _update_vcard_hash(self):
+    def _calculate_vcard_id(self):
         logger.debug("updating vcard hash")
         vcard = yield from self._vcard.get_vcard()
+        logger.debug("%s", vcard)
         photo = vcard.get_photo_data()
-        if photo is not None:
+        if photo is None:
+            # if no photo is set in the vcard set an empty <photo>
+            # element in the update, this means the avatar is disabled
+            return ""
+        else:
             sha1 = hashlib.sha1()
             sha1.update(photo)
-            self._vcard_id = sha1.hexdigest().lower()
+            return sha1.hexdigest().lower()
 
     @service.depsignal(presence.PresenceClient, "on_available")
     def _handle_on_available(self, full_jid, stanza):
-        if full_jid.bare() == self.client.local_jid.bare():
+        if (full_jid.bare() == self.client.local_jid.bare() and
+                full_jid != self.client.local_jid):
             if stanza.xep0153_x is None:
                 self._vcard_ressource_interference.add(full_jid)
             else:
@@ -520,7 +569,7 @@ class AvatarService(service.Service):
         result = collections.defaultdict(lambda: [])
         # note: an empty photo element correctly
         # results in an empty avatar metadata list
-        if notify.photo:
+        if stanza.xep0153_x.photo:
             result[None].append(
                 VCardAvatarDescriptor(
                     remote_jid=jid,
@@ -577,7 +626,8 @@ class AvatarService(service.Service):
     @asyncio.coroutine
     def get_avatar_metadata(self, jid, *, require_fresh=False):
         """
-        Retrieve a list of avatar descriptors for `jid`.
+        Retrieve a mapping from MIME types to avatar descriptors for `jid`.
+        The MIME type is set to :data:`None` if it is not known.
 
         The avatar descriptors are returned as a list of instances of
         :class:`~aioxmpp.avatar.service.AbstractAvatarDescriptor`.
@@ -596,6 +646,8 @@ class AvatarService(service.Service):
             try:
                 logger.debug("trying vCard avatar as fallback for %s", jid)
                 vcard = yield from self._vcard.get_vcard(jid)
+                # XXX: get_photo_data should extract the MIME type as well
+                # here we can know, so we should pass it on
                 photo = vcard.get_photo_data()
                 if photo is not None:
                     logger.debug("success vCard avatar as fallback for %s", jid)
@@ -664,8 +716,8 @@ class AvatarService(service.Service):
                                "on_stream_destroyed")
     def handle_stream_destroyed(self, reason):
         # invalidate the cache?
-        self._vcard_ressource_interference = set()
-        self._has_pep_avatar = set()
+        self._vcard_ressource_interference.clear()
+        self._has_pep_avatar.clear()
 
     @asyncio.coroutine
     def publish_avatar_set(self, avatar_set):
@@ -692,9 +744,6 @@ class AvatarService(service.Service):
                     avatar_set.metadata,
                     id_=id_
                 )
-
-            except RuntimeError:
-                pass
             finally:
                 if self._synchronize_vcard:
                     my_vcard = yield from self._vcard.get_vcard()
@@ -717,8 +766,6 @@ class AvatarService(service.Service):
                     namespaces.xep0084_metadata,
                     avatar_xso.Metadata()
                 )
-            except RuntimeError:
-                pass
             finally:
                 if self._synchronize_vcard:
                     my_vcard = yield from self._vcard.get_vcard()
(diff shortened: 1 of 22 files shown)


aioxmpp/avatar/service.py
```diff
@@ -206,7 +211,8 @@ class AbstractAvatarDescriptor:
     """
 
     def __init__(self, remote_jid, mime_type, id_, nbytes, width=None,
```
horazont: Does it make sense to have all the attributes on the **Abstract**AvatarDescriptor by default? I would’ve expected it to define an interface with abstract properties maybe and let subclasses deal with things.

Especially the ``vcard`` which is exclusively used by the ``VCardAvatarDescriptor`` should be moved there.

``image_bytes`` seems to be unused.

aioxmpp/avatar/service.py
```diff
@@ -340,9 +348,32 @@ def get_image_bytes(self):
...
+
+        raise RuntimeError("Avatar image is not set")
+        # url = item.xpath("/ns0:PHOTO/ns0:EXTVAL/text()",
+        #                  namespaces={"ns0": namespaces.xep0054})
+
+
 class AvatarService(service.Service):
-    """
-    Access and publish User Avatars (:xep:`84`).
+    """Access and publish User Avatars (:xep:`84`). Fallback to vCard based
```
horazont: FWIW, I know that PEP8 likes it, I don’t like the style where the comment starts right after ``"""``, and normally we’re not doing that in aioxmpp.

I don’t think it’s an issue though.

aioxmpp/avatar/service.py
```diff
@@ -373,13 +404,31 @@ class AvatarService(service.Service):
...
+       The PEP descriptor for claiming the avatar metadata namespace.
+       The value is a :class:`~aioxmpp.pep.service.RegisteredPEPNode`,
+       whose :attr:`~aioxmpp.pep.service.RegisteredPEPNode.notify`
+       property can be used to disable or enable the notification
+       feature.
+
+    .. autoattribute:: metadata_cache_size
+       :annotation: = 200
+
     """
```
horazont: .. versionchanged for vcard support.

aioxmpp/avatar/service.py
```diff
@@ -391,17 +440,146 @@ class AvatarService(service.Service):
...
+    @metadata_cache_size.setter
+    def metadata_cache_size(self, value):
+        self._metadata_cache.maxsize = value
+
+    @property
+    def synchronize_vcard(self):
+        """
+        Set this property to true to enable publishing the a vCard avatar.
+
+        This property defaults to false. For the setting true to have
```
horazont: Any technical reason to default to false?

aioxmpp/avatar/service.py
```diff
@@ -391,17 +440,146 @@ class AvatarService(service.Service):
...
+        return self._synchronize_vcard
+
+    @synchronize_vcard.setter
+    def synchronize_vcard(self, value):
+        self._synchronize_vcard = bool(value)
+
+    @service.depfilter(aioxmpp.stream.StanzaStream,
+                       "service_outbound_presence_filter")
+    def _attach_vcard_notify_to_presence(self, stanza):
+        if not self._vcard_ressource_interference:
```
horazont: s/ressource/resource/

aioxmpp/avatar/service.py
```diff
@@ -391,17 +440,146 @@ class AvatarService(service.Service):
...
+            stanza.xep0153_x = avatar_xso.VCardTempUpdate(self._vcard_id)
+        return stanza
+
+    def _handle_notify(self, full_jid, stanza):
+        if stanza.xep0153_x is not None:
```
horazont:
```python
if stanza.x0153_x is None:
    return
```

Saves a whole level of indentation on the remaining method. Same for the next ``if``.

aioxmpp/avatar/service.py
```diff
@@ -391,17 +440,146 @@ class AvatarService(service.Service):
...
+    def _calculate_vcard_id(self):
+        logger.debug("updating vcard hash")
+        vcard = yield from self._vcard.get_vcard()
+        logger.debug("%s", vcard)
+        photo = vcard.get_photo_data()
+        if photo is None:
+            # if no photo is set in the vcard set an empty <photo>
+            # element in the update, this means the avatar is disabled
+            return ""
+        else:
```
horazont: ``else`` isn’t needed here.

aioxmpp/avatar/service.py
```diff
@@ -462,39 +641,69 @@ def get_avatar_metadata(self, jid, *, require_fresh=False):
...
-                pass
-
-        with (yield from self._notify_lock):
-            if jid in self._metadata_cache:
-                if require_fresh:
-                    del self._metadata_cache[jid]
-                else:
-                    return self._metadata_cache[jid]
 
+        def try_vcard_fallback():
```
horazont: Please make this a private method instead (it is not using ``nonlocal``).

aioxmpp/avatar/service.py
```diff
@@ -462,39 +641,69 @@ def get_avatar_metadata(self, jid, *, require_fresh=False):
...
+                    metadata[None] = [VCardAvatarDescriptor(
+                        remote_jid=jid,
+                        mime_type=None,
+                        id_=normalize_id(sha1.hexdigest()),
+                        vcard=self._vcard,
+                        nbytes=len(photo),
+                        image_bytes=photo,
+                    )]
+                    return metadata
+            except aioxmpp.XMPPCancelError:
```
horazont: Make the ``try`` block as small as possible and as large as necessary. The ``logger`` call and everything after ``get_vcard`` should be outside of the ``try`` block.

aioxmpp/avatar/service.py
```diff
@@ -462,39 +641,69 @@ def get_avatar_metadata(self, jid, *, require_fresh=False):
...
-                # item-not-found to be equivalent unset avatar
-                if e.condition in (
-                        (namespaces.stanzas, "feature-not-implemented"),
-                        (namespaces.stanzas, "item-not-found")):
+                logger.debug("trying vCard avatar as fallback for %s", jid)
+                vcard = yield from self._vcard.get_vcard(jid)
+                # XXX: get_photo_data should extract the MIME type as well
+                # here we can know, so we should pass it on
+                photo = vcard.get_photo_data()
+                if photo is not None:
```
horazont:
```python
if photo is None:
    return None
```

aioxmpp/avatar/service.py
```diff
@@ -462,39 +641,69 @@ def get_avatar_metadata(self, jid, *, require_fresh=False):
...
+        if jid in self._metadata_cache:
+            if require_fresh:
+                del self._metadata_cache[jid]
             else:
-                metadata = self._cook_metadata(jid, metadata_raw.payload.items)
+                return self._metadata_cache[jid]
 
-            self._metadata_cache[jid] = metadata
-            return metadata
+        try:
```
horazont: I think refactoring the PEP part into a private method analogous to the ``try_vcard_avatar`` method can be considered (but I haven’t looked at the details yet).

aioxmpp/avatar/service.py
```diff
@@ -503,41 +712,12 @@ def subscribe(self, jid):
...
-            raise_exception()
-
     @aioxmpp.service.depsignal(aioxmpp.stream.StanzaStream,
                                "on_stream_destroyed")
     def handle_stream_destroyed(self, reason):
-        # invalidate the cache
-        self._has_pep = None
+        # invalidate the cache?
+        self._vcard_ressource_interference.clear()
+        self._has_pep_avatar.clear()
```
horazont: Metadata cache?

aioxmpp/avatar/service.py
```diff
@@ -549,24 +729,28 @@ def publish_avatar_set(self, avatar_set):
...
-                avatar_set.metadata,
-                id_=id_
-            )
+                yield from self._pep.publish(
+                    namespaces.xep0084_metadata,
+                    avatar_set.metadata,
+                    id_=id_
+                )
+            finally:
+                if self._synchronize_vcard:
```
horazont: I don’t think we should try to push to the vcard if the push to the PEP failed.

aioxmpp/avatar/service.py
```diff
@@ -575,11 +759,17 @@ def disable_avatar(self):
...
-                namespaces.xep0084_metadata,
-                avatar_xso.Metadata()
-            )
+            try:
+                yield from self._pep.publish(
+                    namespaces.xep0084_metadata,
+                    avatar_xso.Metadata()
+                )
+            finally:
+                if self._synchronize_vcard:
```
horazont: Here, on the other hand, it makes sense to also try to push the vcard if the other thing failed.

I would suggest to use asyncio.gather or something like that, though, because that will (a) parallelize the request and (b) also ensure that both requests are issued, no matter the exceptions.

aioxmpp/avatar/xso.py
```diff
@@ -24,9 +24,32 @@
...
+
 namespaces.xep0084_data = "urn:xmpp:avatar:data"
 namespaces.xep0084_metadata = "urn:xmpp:avatar:metadata"
 
+namespaces.xep0153 = "vcard-temp:x:update"
+
+
+class VCardTempUpdate(xso.XSO):
+    """
+    The vcard temp update note as per :xep:`0153`
```
horazont: s/temp //, it isn’t useful information to the reader.

horazont:
Regarding the races, I think we can assume that push metadata is always the best we get, so it should win all races (1). Only in the cases where we either require fresh or where we don’t get pushes at all we need to do an explicit get.

In light of this, how about this solution for the races:

1. Have two metadata dicts, one with explicitly fetched and one with push metadata.
2. In ``get_avatar_metadata(require_fresh=False)``:

    ```python
    return self._push_metadata.get(jid, self._poll_metadata.get(jid, (yield from do_poll())))
    ```

3. In ``get_avatar_metadata(require_fresh=True)``:

    ```python
    self._push_metadata.pop(jid, None)   # so that subsequent calls to get_avatar_metadata without push inbetween get the just fetched data
    self._poll_metadata.pop(jid, None)
    metadata = yield from do_poll()
    self._poll_metadata[jid] = metadata
    ```
4. In push handlers:

    ```python
    self._poll_metadata.pop(jid, None)  # we don’t need to save polled data if we get pushes
    self._push_metadata[jid] = new_metadata
    ```

The only remaining races I see here are between two ``get_avatar_metadata`` calls for the same ``jid`` which both end up to execute ``do_poll``. Those could be solved by storing ``asyncio.Future`` objects in ``_push_metadata``, like we do in ``query_info`` in ``DiscoClient``, but I’m not sure if it’s worth the trouble.

I think we should defer handling of that race until we have a generic mechanism factored out of ``DiscoClient``.


---
(1): Even if the race is like this:
1. Peer pushes data A
2. ``get_avatar_metadata(require_fresh=True)`` sends the IQ request
3. Push update for data A seen on-stream
4. Peer pushes data B
5. IQ request arrives at pub-sub service, response en-route
6. IQ response received
7. Super-weird coroutine order leads to the IQ response being processed before the push update for data A

it is okay if the push update for data A wins because a push update for data B will be received eventually.

aioxmpp/avatar/service.py
sebastianriese: I think this is the correct approach.

aioxmpp/avatar/service.py
sebastianriese: I believe we discussed this issue before and agreed not to wipe the cache, but I am still not sure which behaviour is correct.

aioxmpp/avatar/xso.py
sebastianriese: and s/note/node/

aioxmpp/avatar/service.py
sebastianriese: Yes, I think using gather is what we want.

aioxmpp/avatar/service.py
sebastianriese: Image bytes is used in the case that the cache is empty and we call get_avatar_metadata which gets the vcard. To prevent another roundtrip we set the image bytes property.

The interface where abstract avatar has all the properties comes from the times when we only supported PEP avatars (where this made much more sense). I think I will have to break the interface slightly, to improve this, but since the constructors of the avatar descriptors are not part of the public interface this should not matter.

horazont: Ah, right, I missed that somehow (and I assumed that ``_vcard`` referred to the actual vcard). Nevermind then.

(Even though PEP 484 type annotations would be great here.)

aioxmpp/avatar/service.py
sebastianriese: What if our server only support vcard?

aioxmpp/avatar/service.py
sebastianriese: I reconfigured my editor to handle this in the preferred (symmetric) style for aioxmpp.

aioxmpp/avatar/service.py
sebastianriese: Okay, gather could work with `return_exceptions = True` which is terrible in-band signalling in my opinion ... otherwise gather raises as soon as the first exception happens (so we would get into trouble if only one of the protocols were supported.

aioxmpp/avatar/service.py
sebastianriese: Meh, for me this is a deep seated German-induced English-bug.

horazont: So the same like I have for "address". Took me years to figure that out.

aioxmpp/avatar/service.py
horazont: Right, good point. In that case, we need to try both, so using asyncio.gather/asyncio.wait seems reasonable.

aioxmpp/avatar/service.py
horazont: We don’t need the return values in this case, so it shouldn’t matter.

(If you care about this strongly, you could fake things with:
```python
done, pending = asyncio.wait([coro1, coro2], return_when=asyncio.ALL_COMPLETED)
for fut in done:
    fut.exception() or fut.result()  # needed to avoid warnings
```
… which funnily enough does exactly the same thing like asyncio.gather, probably)

aioxmpp/avatar/service.py
sebastianriese: Primarily aesthetic reasons (don't actively use the legacy protocol by default)

horazont: Ejabberd would have to be able to work with that, it’s not the clients fault in that case (if they need special casing, they should negotiate that feature).

I don’t have strong opinions either way, in any case we’ll just enable it in JabberCat.

aioxmpp/avatar/service.py
```diff
@@ -748,33 +807,107 @@ def publish_avatar_set(self, avatar_set):
...
-                    yield from self._presence_server.resend_presence()
+                done = True
+
+            if self._synchronize_vcard:
+                my_vcard = yield from self._vcard.get_vcard()
+                my_vcard.set_photo_data("image/png",
+                                        avatar_set.image_bytes)
+                self._vcard_id = avatar_set.png_id
+                yield from self._vcard.set_vcard(my_vcard)
+                yield from self._presence_server.resend_presence()
```
horazont: ``resend_presence`` is not a coroutine. It can return a (awaitable) stanza token, but it may also return None.

I also don’t think that waiting for the resend is necessary.

aioxmpp/avatar/service.py
```diff
@@ -748,33 +807,107 @@ def publish_avatar_set(self, avatar_set):
...
+                "failed to publish avatar: no protocol available"
+            )
+
+    @asyncio.coroutine
+    def _disable_vcard_avatar(self):
+        my_vcard = yield from self._vcard.get_vcard()
+        my_vcard.clear_photo_data()
+        self._vcard_id = ""
+        yield from self._vcard.set_vcard(my_vcard)
+        yield from self._presence_server.resend_presence()
```
horazont: See above.

aioxmpp/avatar/service.py
```diff
@@ -748,33 +807,107 @@ def publish_avatar_set(self, avatar_set):
...
+                ))
+
+            done, _ = yield from asyncio.wait(todo)
+            exceptions = []
+            for fut in done:
+                if fut.exception() is not None:
+                    exceptions.append(fut.exception())
+                fut.result() # to avoid warnings
+            if exceptions:
+                raise RuntimeError from exceptions[0]
```
horazont: I really wish python had a built-in way to re-raise multiple exceptions :(.

In any case, this exception deserves a text. I’d suggest ``"Failed to disable avatar ({})".format(", ".join(map(repr, exceptions)))``

aioxmpp/avatar/service.py
```diff
@@ -748,33 +807,107 @@ def publish_avatar_set(self, avatar_set):
...
+                todo.append(_wipe_pep_avatar())
+
+            done, _ = yield from asyncio.wait(todo)
+            exceptions = []
+            for fut in done:
+                if fut.exception() is not None:
+                    exceptions.append(fut.exception())
+                fut.result() # to avoid warnings
+            if exceptions:
+                raise RuntimeError from exceptions[0]
```
horazont: See above.

aioxmpp/avatar/service.py
```diff
@@ -748,33 +807,107 @@ def publish_avatar_set(self, avatar_set):
...
-                    yield from self._vcard.set_vcard(my_vcard)
-                    yield from self._presence_server.resend_presence()
+                ))
+
+            done, _ = yield from asyncio.wait(todo)
+            exceptions = []
+            for fut in done:
+                if fut.exception() is not None:
+                    exceptions.append(fut.exception())
+                fut.result() # to avoid warnings
```
horazont: fut.result() raises if the future raised an exception.

aioxmpp/avatar/service.py
```diff
@@ -748,33 +807,107 @@ def publish_avatar_set(self, avatar_set):
...
+
+            if (yield from self._pep.available()):
+                todo.append(_wipe_pep_avatar())
+
+            done, _ = yield from asyncio.wait(todo)
+            exceptions = []
+            for fut in done:
+                if fut.exception() is not None:
+                    exceptions.append(fut.exception())
+                fut.result() # to avoid warnings
```
horazont: See above.

aioxmpp/avatar/service.py
```diff
@@ -748,33 +807,107 @@ def publish_avatar_set(self, avatar_set):
...
+            todo = []
+            if self._synchronize_vcard:
+                todo.append(self._disable_vcard_avatar())
+
+            if (yield from self._pep.available()):
+                todo.append(_wipe_pep_avatar())
+
+            done, _ = yield from asyncio.wait(todo)
+            exceptions = []
+            for fut in done:
```
horazont: Maybe refactor this handling in its own function?

examples/retrieve_avatar.py
```diff
@@ -74,14 +74,14 @@ def run_simple_example(self):
...
-            if metadatum.has_image_data_in_pubsub:
+        for metadatum in metadata:
+            if metadatum.can_get_image_bytes_via_xmpp:
                 image = yield from metadatum.get_image_bytes()
                 with open(self.output_file, "wb") as avatar_image:
                     avatar_image.write(image)
                 return
 
-        print("retrieving avatar failed: no avatar in pubsub")
+        print("retrieving avatar failed: no png-avatar in pubsub")
```
horazont: Does this code check for PNG? (The XEP *says* that the PubSub data avatars must be PNG, but …)

horazont:
Erm, this was supposed to be "Request changes", sorry.

aioxmpp/avatar/service.py
sebastianriese: oops, should have read the docs better, good catch!

aioxmpp/avatar/service.py
sebastianriese: yup, there *should* be an else ...

examples/retrieve_avatar.py
sebastianriese: Ow, this is a left-over from an earlier version, where I tested for the mime-type, I stopped this due to the possibility of `mime_type is None`. It should not even say pubsub, as it falls back to vCard.

PR #167, opened 2017-12-22
Implement XEP-0092 (Software Version)

horazont:
Fixes #167.

As the title states :).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,6 +99,10 @@ Dependencies
 
   __ https://github.com/horazont/aioopenssl
 
+* `typing`__ (Python < 3.5 only)
+
+  __ https://pypi.python.org/pypi/typing
+
 Contributing
 ============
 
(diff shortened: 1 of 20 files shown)


sebastianriese:
If I read the code right, we allow to del the "name" and "version" attributes we send out, this is prohibited by the specification: <https://xmpp.org/extensions/xep-0092.html#protocol>. Note, "name" and "version" are required only "os" is optional.

sebastianriese:
A mechanism to specify a filter policy should be easy to provide (e.g. as a hook of a "policy service"), we could simply provide an "allow all" policy by default (which would be easy to implement).

(Or we do this with the postponed firewall service, which should be freely configurable with some kind of pattern matching).

horazont:
> If I read the code right, we allow to del the "name" and "version" attributes we send out, this is prohibited by the specification: https://xmpp.org/extensions/xep-0092.html#protocol. Note, "name" and "version" are required only "os" is optional.

Good catch! Filing an issue for that.

> A mechanism to specify a filter policy should be easy to provide (e.g. as a hook of a "policy service"), we could simply provide an "allow all" policy by default (which would be easy to implement).

I’m not sure if this is super-relevant, so I propose to again postpone this to a future firewall/policy service.

PR #171, opened 2018-02-20
Ensure that presence is resent after vcard hash update

horazont:


sebastianriese:
This sends the presence too often. I believe, the bug is only triggered if we receive a cached presence from the server and then send our own presence before we are finished processing the vcard ID. Then the server will cache our presence with an empty update element, thereby breaking the avatar for all peers that have not already cached our avatar. If we just receive a presence update from the peer and the hash matches the hash in the vcard we should not send a new presence (since this would be unnecessary additional traffic). Actually, I just noticed, we violate the XEP by not sending an new presence with *empty* `<update>` element *before* we continue to retrieve the vcard and calculate the new hash (see <https://xmpp.org/extensions/xep-0153.html#bizrules-reset>), but I am not sure the prescribed behaviour is really a good idea. Perhaps, we should resend the presence if, and only if, the newly calculated hash does not match the hash in the previous presence (and prehaps additionally if our previously cached hash was None) – but note that this behaviour may trigger a loop (if one client's hash calculation or encoding is buggy and they disagree what the hash of the same image is).

sebastianriese:
Okay, I thought about it, and sending the empty update-element immediately and sending the new hash with all later presence updates is the fastest route to a consistent state, since the last presence stanza from each resource is cached by the server and sent out in response to presence probes, this means that setting the update element empty, will make the state consistent: The resource that uploaded the new avatar will send out the presence with the new hash, this presence is cached and sent out as part of the reply to presence probes, if all other resources set the update element empty (which means it is ignored in the processing), means that (newly connecting) peers will observe a consistent state, since only in the presence of one resource hash is set and that hash corresponds to the avatar image in the vCard. Further, this consistent state is reached faster than in the case where all other resources first download the vCard and calculate the hash before they send a new presence with the new hash.

sebastianriese:
The new commits should do the right thing (modulo perhaps sending too many empty presence stanzas when receiving the cached presence for the other ressources).

PR #175, opened 2018-02-22
Take care of the startup races

### LINKED ISSUE #173: Do not allow forwarding of application stanzas until resource binding has completed
Currently, when an application enqueues a stanza before the stream has first been established, the stanza is sent *before* the resource binding takes place. This can lead to all sorts of funny issues:

* messages will be lost or errored back
* IQ requests will fail with not-authorized
* presence may be errored back or be lost

… much of this (except IQs) happens without the application being the wiser.

The stream should not accept stanzas from services or application code until it has been fully negotiated.

### LINKED ISSUE #174: Ensure that stream management is enabled synchronously
Apparently, it is currently possible to make aioxmpp fail by sending a ``<{sm}r/>`` right after enabling stream management. Of course, this is nonsensical to do normally, but in combination with #173, it is possible that we’re sending stanzas right after the ``<{sm}enable/>``, prompting the peer to reply with ``<{sm}enabled/>``, some replies and then ``<{sm}r/>``. 

In this case, it is apparently possible (I’ve got logs to prove it!) that the SM handlers are not attached to the XML stream yet, thus making the stream fail with ``unsupported-stanza-type``.

horazont:
Fixes #174, fixes #173.

This does quite a bit of things:

* fixes the Stream Management race by doing things right
* deprecates the ``send()`` and ``enqueue()`` methods on the ``StanzaStream`` by moving them to the ``Client``
* makes ``enqueue()`` raise if the stream is not established yet
* makes ``send()`` block until the stream is established (and raise if the stream fails to establish)

TODO:

* [x] fix all the deprecated uses in our services and tests
* [x] fix the deprecated uses in the docs
* [x] fix the deprecated uses in the examples

(The things on the TODO should not affect the review, so feel free to start.)

### DIFF
--- a/aioxmpp/node.py
+++ b/aioxmpp/node.py
@@ -567,6 +567,12 @@ class Client:
        when :attr:`before_stream_established` fires, the information is
        up-to-date.
 
+    Sending stanzas:
+
+    .. automethod:: send
+
+    .. automethod:: enqueue
+
     Configuration of exponential backoff for reconnects:
 
     .. attribute:: backoff_start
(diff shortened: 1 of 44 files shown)


aioxmpp/node.py
```diff
@@ -1132,6 +1167,154 @@ def connected(self, *, presence=structs.PresenceState(False), **kwargs):
...
+            * This method was moved from
+              :meth:`aioxmpp.stream.StanzaStream.send`.
+
+        .. versionchanged:: 0.9
+
+            The `cb` argument was added.
+
+        .. versionadded:: 0.8
+        """
+        stopped_fut = self.on_stopped.future()
```
sebastianriese: This requires an extra scheduling cycle for every send! It may be better to check the state of the established_event and only construct the futures and wait for them it if established_event is not set!

horazont: Good point!

PR #181, opened 2018-03-19
Implement ALPN support

sebastianriese:
Not sure if it is correct, that we also send ALPN when we do STARTTLS, if this should be an issue I will fix it.


horazont:
Don’t we have tests for ``default_ssl_context``?

We should add some at some point. Ideally neat end-to-end-like tests where we try to connect to e.g. SSLv3 hosts and make sure that that fails.

PR #183, opened 2018-03-21
connector/security_layer: Move ALPN support to connector

### LINKED ISSUE #170: XEP-0368: Implement ALPN support

sebastianriese:
Fixes #170.

* ALPN support now lives in the connector module

* improved checking for ALPN availablity (pyOpenSSL raises
  NotImplementedError if the underlying OpenSSL does not
  support ALPN)

* warnings if ALPN is required but not available

NOTE: e2e tests for the ALPN support in the connector would be nice.

### DIFF
--- a/tests/test_connector.py
+++ b/tests/test_connector.py
@@ -28,7 +28,6 @@ import unittest.mock
 import aioxmpp.connector as connector
 import aioxmpp.errors as errors
 import aioxmpp.nonza as nonza
-import aioxmpp.security_layer as security_layer
 
 from aioxmpp.utils import namespaces
 
@@ -189,7 +188,8 @@ class TestSTARTTLSConnector(unittest.TestCase):
                 ),
                 unittest.mock.call.protocol.starttls(
                     ssl_context=unittest.mock.sentinel.ssl_context,
-                    post_handshake_callback=base.certificate_verifier.post_handshake,
+                    post_handshake_callback=
+                        base.certificate_verifier.post_handshake,
                 ),
                 unittest.mock.call.reset_stream_and_get_features(
                     base.protocol,
@@ -437,7 +437,8 @@ class TestSTARTTLSConnector(unittest.TestCase):
                 ),
                 unittest.mock.call.protocol.starttls(
                     ssl_context=unittest.mock.sentinel.ssl_context,
-                    post_handshake_callback=base.certificate_verifier.post_handshake,
+                    post_handshake_callback=
+                        base.certificate_verifier.post_handshake,
                 ),
                 unittest.mock.call.reset_stream_and_get_features(
                     base.protocol,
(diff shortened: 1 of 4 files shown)


tests/test_connector.py
```diff
@@ -1216,6 +1221,67 @@ def capture_future(*args, features_future=None, **kwargs):
             )
         )
 
+        # test for correct generation of warnings if set_alpn_proto is
```
horazont: This should be a separate test case. If needed, you can refactor the context factory out of the connect method into a method of the object. That might be saner anyways. Allows to refactor the test for that context factory wrapper out into separate tests, too. Then we only need to test against correct use of functools partial, if that’s even needed.

tests/test_connector.py
```diff
@@ -1116,8 +1117,10 @@ def capture_future(*args, features_future=None, **kwargs):
         base.certificate_verifier.pre_handshake = CoroutineMock()
         base.metadata.certificate_verifier_factory.return_value = \
             base.certificate_verifier
-        base.metadata.ssl_context_factory.return_value = \
-            unittest.mock.sentinel.ssl_context
+        # base.metadata.ssl_context_factory.return_value = \
+        #     unittest.mock.sentinel.ssl_context
```
horazont: Delete those lines.

tests/test_connector.py
sebastianriese: Yes, factoring out the mock setup sounds right. I did it this way, because I considered avoiding the code duplication to be more important than having separate test cases, don't ask why I did not think of factoring the setup out.

tests/test_connector.py
sebastianriese: Oops, why are those lines still there.

tests/test_connector.py
```diff
@@ -1195,16 +1201,18 @@ def capture_future(*args, features_future=None, **kwargs):
             base.mock_calls,
             [
                 unittest.mock.call.metadata.ssl_context_factory(),
+                unittest.mock.call.metadata.
+                    ssl_context_factory().set_alpn_protos([b"xmpp-client"]),
```
horazont: Do you have objections to changing this testcase to test that _context_factory_factory is invoked and used instead of testing that it does something similar to what the context_factory created by the _context_factory_factory does?

sebastianriese: And test what `_context_factory_factory(...)(...)` does separately? I agree that this would be a cleaner separation of tests. I am a bit reluctant because the change might introduce bugs in the test and thereby mask a regression due to the change in connector.py (note, how the test did not change apart from the added call to `set_alpn_protos`! So it demonstrates that the behaviour before and after the change is the same with respect to the mocked objects.).

sebastianriese: Okay, the changes where minimal enough so that I am comfortable with the change (I could simply move the call-sequence assertion to the new test). And once again I have to remark: `unittest.mock` is pure joy ;).

PR #186, opened 2018-03-27
misc: Add XSO definitions for XEP-0379

### LINKED ISSUE #104: XEP-0379 (Pre-Authenticated Roster Subscription) support
[XEP-0379 (Pre-Authenticated Roster Subscription)](https://xmpp.org/extensions/xep-0379.html)

sebastianriese:
This adds what is necessary on the aioxmpp side to support XEP 379 and therefore closes #104.

### DIFF
--- a/aioxmpp/misc/__init__.py
+++ b/aioxmpp/misc/__init__.py
@@ -64,6 +64,14 @@ Chat Markers (:xep:`333`)
 
 .. attribute:: aioxmpp.Message.xep0333_marker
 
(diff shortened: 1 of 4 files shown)


sebastianriese:
Hmm, presence seems to have a `ChildMap` `ext`, should I use that instead of monkey patching the property?

horazont:
@sebastianriese Please don’t. That is legacy from before monkey patching descriptors.

sebastianriese:
Good to know. Monkey patching seems to be more reasonable anyway (from an API point of view).

PR #191, opened 2018-05-18
Improve message tracking and delivery receipts

### LINKED ISSUE #166: tracking: Make sure that delivery reciepts can trump error replies
Currently, the transition from ERROR to anything else is forbidden. We need to fix this to allow message delivery reciepts to trump errors from, e.g., a killed XEP-0198 session.

horazont:
* Include documentation for the mdr (XEP-0184) module and make it more usable.
* Add e2e tests for the mdr module.
* Fix #166.

### DIFF
--- a/aioxmpp/__init__.py
+++ b/aioxmpp/__init__.py
@@ -134,6 +134,7 @@ from .ping import PingService  # NOQA
 from .pep import PEPClient  # NOQA
 from .bookmarks import BookmarkClient  # NOQA
 from .version import VersionServer  # NOQA
+from .mdr import DeliveryReceiptsService # NOQA
 
 
 def set_strict_mode():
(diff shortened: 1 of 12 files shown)


aioxmpp/__init__.py
```diff
@@ -134,6 +134,7 @@
 from .pep import PEPClient  # NOQA
 from .bookmarks import BookmarkClient  # NOQA
 from .version import VersionServer  # NOQA
+from .mdr import DeliveryReceiptsService # NOQA
```
sebastianriese: PEP8 violation (there need to be two spaces before `#`).

docs/glossary.rst
```diff
@@ -0,0 +1,33 @@
+Glossary
```
sebastianriese: Does this file really belong to this PR?

horazont: It was kinda "created" while writing the documentation. It is not 100% right here, but I figured "eh". Perfect is the enemy of good.

aioxmpp/__init__.py
horazont: good catch

PR #196, opened 2018-05-19
Remove libxml2 dependency

### LINKED ISSUE #193: Loading xml2 lib does not work in MSYS2/Windows
if libxml2 is installed on Windows via MSYS2 the .dll is called `libxml2-2.dll`

So the code would have to be adapted

xml.py
```python
if os.name == 'nt':
    libxml2 = ctypes.cdll.LoadLibrary(
        ctypes.util.find_library("libxml2-2.dll"),
    )
```

i think over time you will have to test for multiple names, depending on where and how lxml is installed

### LINKED ISSUE #195: Get rid of libxml dependency
We currently load libxml2 for a validation function. Ideally, we’d re-write this function in pure python to get rid of the burden to find libxml2 (despite it’ll be loaded anyways because we also depend on lxml).

**up-for-grabs note:** Please see the README and join our chat room if you want help or guidelines on how to implement this.

horazont:
Fixes #193 and fixes #195.

This has the downside of slowing down XML serialisation. There are opportunities for future optimization, but we’ll just do that at some other time.

### DIFF
--- a/aioxmpp/xml.py
+++ b/aioxmpp/xml.py
@@ -75,50 +75,42 @@ from . import errors, structs, xso
 from .utils import namespaces
 
 
-_NAME_START_CHAR = (
-    set([
-        ":",
-        "_",
-    ]) |
-    set(chr(ch) for ch in range(ord("a"), ord("z")+1)) |
-    set(chr(ch) for ch in range(ord("A"), ord("Z")+1)) |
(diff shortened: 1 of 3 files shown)


horazont:
I see your point, but given that I generated the test case while we were still using the libxml2 version, I’m confident that this is not needed.

sebastianriese:
My proposal is more of an aesthetic than a technical nature (and can safely be ignored):

PR #199, opened 2018-05-21
Update setup.py

zwerheim:


horazont:
Out of curiosity, what was this about?

PR #203, opened 2018-06-04
Fixes to the Data Forms implementation

### LINKED ISSUE #200: XEP-0004: Do not perform validation based on field type during reception
Validation is more tricky than that, because the presence of the ``field@type`` is not guaranteed: It is a SHOULD for ``form@type="form"`` and a MAY for all other types.

### LINKED ISSUE #201: XEP-0004: Allow arbitrary form types when loading forms from XSO
This is needed for XEP-0128.

### LINKED ISSUE #202: XEP-0004: Do not treat absent field type as "text-single"
"text-single" is only the default for **unknown**, not for **absent** field types.

horazont:
Specifically, this fixes #200, fixes #201, fixes #202 and addresses other issues.

### DIFF
--- a/aioxmpp/forms/fields.py
+++ b/aioxmpp/forms/fields.py
@@ -357,6 +357,9 @@ class BoundSingleValueField(BoundField):
         except AttributeError:
             value = self._field.default()
 
+        if value is None:
+            return result
+
         result.values[:] = [
             self.field.type_.format(value)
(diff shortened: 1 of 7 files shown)


horazont:
It passes tests, it works in the wild (in fact, I run muclumbus on this branch at the moment), gonna merge this :)

PR #207, opened 2018-06-05
Various improvements to MUC

horazont:
* Support for voice requests
* Harden tests against room locking
* Use enums for status codes.

More to come. Early reviews appreciated, but don’t waste your time. This PR is against the feature/form-fixes branch for readability, but I will change that once feature/form-fixes is merged.


horazont:
I have no idea why github doesn’t realize this is merged O_o.

PR #211, opened 2018-06-08
XEP-0077 In-Band Registration added.

Gersiete:
e2e tests are currently work in progress.

### DIFF
--- a/aioxmpp/ibr/service.py
+++ b/aioxmpp/ibr/service.py
@@ -25,23 +25,121 @@ import asyncio
 import aioxmpp
 import logging
 
+from aioxmpp.service import Service
 from . import xso
 
 
 logger = logging.getLogger(__name__)
 
 
-class RegistrationService:
+@asyncio.coroutine
+def get_registration_fields(xmlstream,
+                            timeout=60,
+                            ):
+    """
+    A query is sent to the server to obtain the fields that need to be
+    filled to register with the server.
+
+    :param xmlstream: Specifies the stream connected to the server where
+                      the account will be created.
+    :type xmlstream: :class:`aioxmpp.protocol.XMLStream`
+
+    :param timeout: Maximum time in seconds to wait for an IQ response, or
+                    :data:`None` to disable the timeout.
+    :type timeout: :class:`~numbers.Real` or :data:`None`
+
+    :return: :attr:`list`
+    """
+
+    iq = aioxmpp.IQ(
+        to=aioxmpp.JID.fromstr(xmlstream._to),
+        type_=aioxmpp.IQType.GET,
+        payload=xso.Query()
+    )
+    iq.autoset_id()
+
+    reply = yield from aioxmpp.protocol.send_and_wait_for(xmlstream,
+                                                          [iq],
+                                                          [aioxmpp.IQ],
+                                                          timeout=timeout)
+    return reply.payload
+
+
+@asyncio.coroutine
+def register(query_xso,
+             xmlstream,
+             timeout=60,
+             ):
+    """
+    Create a new account on the server.
+
+    :param query_xso: XSO with the information needed for the registration.
+    :type query_xso: :class:`xso.Query`
+
+    :param xmlstream: Specifies the stream connected to the server where
+                      the account will be created.
+    :type xmlstream: :class:`aioxmpp.protocol.XMLStream`
+
+    :param timeout: Maximum time in seconds to wait for an IQ response, or
+                    :data:`None` to disable the timeout.
+    :type timeout: :class:`~numbers.Real` or :data:`None`
+    """
+    iq = aioxmpp.IQ(
+        to=aioxmpp.JID.fromstr(xmlstream._to),
+        type_=aioxmpp.IQType.SET,
+        payload=query_xso
+    )
+    iq.autoset_id()
+
+    yield from aioxmpp.protocol.send_and_wait_for(xmlstream,
+                                                  [iq],
+                                                  [aioxmpp.IQ],
+                                                  timeout=timeout)
+
+
+def get_used_fields(payload):
+    """
+    Get a list containing the names of the fields that are used in the
+    xso.Query.
+
+    :param payload: Query object o be
+    :type payload: :class:`xso.Query`
+    :return: :attr:`list`
+    """
+    return [a for a in dir(payload)
+            if isinstance(getattr(payload, a), str) and
+            not a.startswith('__')]
+
+
+def get_query_xso(username, password, aux_fields=None):
+    """
+    Get an xso.Query object with the info provided in he parameters.
+
+    :param username: Username of the query
+    :type username: :class:`str`
+    :param password: Password of the query.
+    :type password: :class:`str`
+    :param aux_fields: Auxiliary fields in case additional info is needed.
+    :type aux_fields: :class:`dict`
+    :return: :class:`xso.Query`
+    """
+    query = xso.Query()
+    query.username = username
+    query.password = password
+
+    if aux_fields is not None:
+        for key, value in aux_fields.items():
+            setattr(query, key, value)
+
+    return query
+
+
+class RegistrationService(Service):
     """
     Service implementing XMPP In-Band Registration(:xep:`0077`).
 
-    This service implements the response to XMPP Pings and provides a method
-    to send pings.
-
-    This class doesn't inherit from the Service class because here are methods
-    that can't have a connected client in order to run. For example, a new
-    account should sent the needed stanza without being authenticated,
-    otherwise it would make no sense.
+    This 'service' implements the possibility for an entity to register with a
+    XMPP server, cancel an existing registration, or change a password.
 
     .. automethod:: get_client_info
 
@@ -56,69 +154,54 @@ class RegistrationService:
     """
 
     @asyncio.coroutine
-    def get_client_info(self, client):
+    def get_client_info(self):
         """
-        A query is sent to the server o obtain the client's data stored at the
+        A query is sent to the server to obtain the client's data stored at the
         server.
 
-        :param client: Client from which the query is sent.
-        :type client: :class:`aioxmpp.Client`
-
-        :return: IQ response :attr:`~.IQ.payload` or :data:`None`
+        :return: :class:`xso.Query`
         """
         iq = aioxmpp.IQ(
-            to=client.local_jid.bare().replace(localpart=None),
+            to=self.client.local_jid.bare().replace(localpart=None),
             type_=aioxmpp.IQType.GET,
             payload=xso.Query()
         )
 
-        return (yield from client.send(iq))
+        reply = (yield from self.client.send(iq))
+        return reply
 
     @asyncio.coroutine
     def change_pass(self,
-                    client,
                     new_pass,
-                    aux_fields=None):
+                    old_pass=None):
         """
         Change the client password for 'new_pass'.
-
-        :param client: Client who wants to change the password.
-        :type client: :class:`aioxmpp.Client`
 
         :param new_pass: New password of the client.
         :type new_pass: :class:`str`
 
-        :param aux_fields: Auxiliary fields in case additional info is needed.
-        :type aux_fields: :class:`dict`
+        :param old_pass: Old password of the client.
+        :type old_pass: :class:`str`
 
-        :return:`None`
-
-        If the query fails for whatever reason, it will raise an exception
-        from inside client.send(iq).
         """
         iq = aioxmpp.IQ(
-            to=client.local_jid.bare().replace(localpart=None),
+            to=self.client.local_jid.bare().replace(localpart=None),
             type_=aioxmpp.IQType.SET,
             payload=xso.Query()
         )
 
-        iq.payload.username = client.local_jid.localpart
+        iq.payload.username = self.client.local_jid.localpart
         iq.payload.password = new_pass
 
-        if aux_fields is not None:
-            for key, value in aux_fields.items():
-                setattr(iq.payload, key, value)
-        yield from client.send(iq)
+        if old_pass:
+            iq.payload.old_password = old_pass
+
+        yield from self.client.send(iq)
 
     @asyncio.coroutine
-    def cancel_registration(self, client):
+    def cancel_registration(self):
         """
         Cancels the currents client's account with the server.
-
-        :param client: Client who wants to cancel.
-        :type client: :class:`aioxmpp.Client`
-
-        :return:`None`
 
         Even if the cancelation is succesful, this method will raise an
         exception due to he account no longer exists for the server, so the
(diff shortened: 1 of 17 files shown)


setup.py
```diff
@@ -40,7 +40,7 @@
     'babel~=2.3',
     'dnspython~=1.0',
     'lxml~=3.6',
-    'multidict~=2.0',
+    'multidict<5,>=4.0',
```
sebastianriese: Could the same not be achieved with `~=4.0`?

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+        iq.payload.password = password
+
+        if aux_fields is not None:
+            for key, value in aux_fields.items():
+                setattr(iq.payload, key, value)
+
+        loop = asyncio.get_event_loop()
+        metadata = aioxmpp.make_security_layer(password, no_verify=no_verify)
+
+        try:
```
sebastianriese: There is no need for a try-except block if the except block simply re-raises the exception.

horazont:
Thank you for the PR!

I will take a first look this weekend (I won’t get around to do it today, unfortunately). On the first glance I found a few things I’m not 100% happy with, but no big issues yet. :+1:

Regarding end-to-end-tests: This will probably require changes to the test server configurations because they’re currently not configured to allow or even support registration. We will have to work out a solution for that (which might be a bit tricky).

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+            to=jid.bare().replace(localpart=None),
+            type_=aioxmpp.IQType.SET,
+            payload=xso.Query()
+        )
+        iq.autoset_id()
+
+        iq.payload.username = jid.localpart
+        iq.payload.password = password
+
+        if aux_fields is not None:
```
sebastianriese: Perhaps expose the XSO query class instead (for example by having an advanced method that takes an XSO instead of `jid` and `password`). Exposing XSOs across module boundaries is not per se considered bad style in aioxmpp. I don't think this solution is particularly bad, but I like to limit the use of `setattr` and the likes to the absolutely necessary.

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+            to=peer_jid.bare().replace(localpart=None),
+            type_=aioxmpp.IQType.GET,
+            payload=xso.Query()
+        )
+        iq.autoset_id()
+
+        loop = asyncio.get_event_loop()
+        metadata = aioxmpp.make_security_layer('', no_verify=no_verify)
+
+        try:
```
sebastianriese: There is no point in a try-except block where the exception is simply re-raised.

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+            reply = yield from self.connect_and_send(iq, loop, metadata, peer_jid, timeout)
+
+            return [a for a in dir(reply.payload)
```
sebastianriese: Perhaps simply return the payload XSO instead instead of decomposing it into a list. If the payload XSO grows a public method later on things get uncomfortable with this code. (The person adding the method might overlook this code and then the method will appear in the dictionary: not what we want).

horazont:
An example would be a great thing. It serves as a quick way to manually test the feature, in case e2e tests can’t be written easily.

sebastianriese:
Sorry for the broken anchors in my review, I still don't get why github thinks that I want context before but not after an anchor when commenting on code. But my comments should be understandable when you look at them in the full diff.

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class RegistrationService:
+    """
+    Service implementing XMPP In-Band Registration(:xep:`0077`).
+
+    This service implements the response to XMPP Pings and provides a method
```
javipalanca: copy-paste typo in doc. It's not the response to XMPP Pings. Implements the In-Band Registration.

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+    .. automethod:: get_registration_fields
+
+    .. automethod:: register
+
+    """
+
+    @asyncio.coroutine
+    def get_client_info(self, client):
+        """
+        A query is sent to the server o obtain the client's data stored at the
```
javipalanca: typo: o obtain -> to obtain

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+            transport, xmlstream, features = yield from conn.connect(
+                loop,
+                metadata,
+                peer_jid.domain,
+                host,
+                port,
+                timeout,
+                base_logger=logger
+            )
+            reply = yield from aioxmpp.send_and_wait_for(
```
horazont: Use ``from ..protocol import send_and_wait_for`` at the top of this module instead of importing ``send_and_wait_for`` into ``aioxmpp``.

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+                xmlstream, [iq], [aioxmpp.IQ]
+            )
+        return reply
```
horazont: This doesn’t properly close the ``xmlstream``. This needs to handle exceptions from ``send_and_wait_for`` with a try-finally block and close the xmlstream properly.

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+
+    @asyncio.coroutine
+    def get_registration_fields(self,
+                                peer_jid,
+                                timeout=60,
+                                no_verify=True
```
horazont: ``no_verify`` must not default to true.

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+            type_=aioxmpp.IQType.SET,
+            payload=xso.Query()
+        )
+
+        iq.payload.remove = True
+        yield from client.send(iq)
+
+    @staticmethod
+    @asyncio.coroutine
+    def connect_and_send(iq, loop, metadata, peer_jid, timeout):
```
horazont: I am not too happy with ``connect_and_send``. This is has extremely high overhead: it does DNS lookups (modulo caching), TCP connect, TLS handshake only to send a single IQ and then abort.

If you look at how aioxmpp core handles establishing a connection (``aioxmpp.node._try_options`` and ``aioxmpp.node.connect_xmlstream``), you will notice that the error handling is also much more complex than what you currently have.

I suggest to, instead of completely re-writing the code which establishes the XML stream yourself, modify ``aioxmpp.node._try_options`` to be able to do what you need. If I’m not mistaken, your requirements are:

1. Get an XMLStream
2. Without authentication (pre-SASL)

Achieving this can be as simple as adding an ``if`` to ``_try_options`` which checks if any SASL providers are given in the security context (``metadata``). If the list is empty, simply skip sasl negotiation. You will have to modify ``node.py`` and write a few simple tests for ``_try_options``, but that’s okay. We are adding a new feature to aioxmpp core after all (which is: a way to handle pre-authentication things).


You can then replace ``connect_and_send`` with ``connect_xmlstream`` with a metadata which has an empty list of ``sasl_providers``. The usage of the IBR API would then look roughly like this:

1. ``_, xmlstream, features = connect_xmlstream()``
2. check if IBR offered in ``features``, otherwise fail early
3. use ``aioxmpp.ibr.get_registration_fields(xmlstream)`` to obtain the fields required
4. fill out the fields
5. invoke ``aioxmpp.ibr.register(xmlstream, fields)``

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+        options = list((
+            yield from aioxmpp.discover_connectors(
+                peer_jid.domain, loop=loop
+            )
+        ))
+        for host, port, conn in options:
```
horazont: The loop as written will attempt to send the IQ to each of the options and return the reply of the last option. This is not correct behaviour.

aioxmpp/ibr/service.py
horazont: @sebastianriese is right that it’s better to return the XSO payload here instead of trying to parse it in the method. Try to keep methods as minimalistic as possible (without over-doing it), this helps with re-usability and testability.

I suggest to add a method (e.g. ``get_used_fields`` or something similar) to the ``Query`` class instead which returns the set of child element names which were given. This can be done for example by iterating over ``Query.CHILD_MAP``. Try this:

```
q = aioxmpp.ibr.xso.Query()
q.username = ""
q.password = ""
print([
    tag
    for tag, descriptor in q.CHILD_MAP.items()
    if descriptor.__get__(q, type(q)) is not None
])
```

This gives you a list of tags of all (child) properties which are not None. You will want to exclude the instructions field from the result of that method.

aioxmpp/ibr/service.py
horazont: This also plays nice with the change suggested to ``get_registration_fields``.

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+        metadata = aioxmpp.make_security_layer('', no_verify=no_verify)
+
+        try:
+            reply = yield from self.connect_and_send(iq, loop, metadata, peer_jid, timeout)
+
+            return [a for a in dir(reply.payload)
+                    if isinstance(getattr(reply.payload, a), str) and
+                    not a.startswith('__')]
+        except Exception as exc:
+            raise exc
```
horazont: The ``get_registration_fields`` method does not at all account for 

* the possibility that the server returns a data form instead of the classic registration fields (XEP-0077 is a historically-grown spec which didn’t age too well, unfortunately).
* the possibility that the server only returns ``<instructions/>`` for an out-of-band registration.

The code should at least detect those cases and raise an error.

By returning the XSO instead of attempting to parse it, it’s however possible to circumvent this necessity.

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+        iq.autoset_id()
+
+        loop = asyncio.get_event_loop()
+        metadata = aioxmpp.make_security_layer('', no_verify=no_verify)
+
+        try:
+            reply = yield from self.connect_and_send(iq, loop, metadata, peer_jid, timeout)
+
+            return [a for a in dir(reply.payload)
+                    if isinstance(getattr(reply.payload, a), str) and
```
horazont: This check can be made more readable by checking against ``is not None`` instead of ``isinstance(…, str)``. The type can only be str or NoneType, and it not being None is what you’re really interested in (because that means it has not been defaulted).

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+        except Exception as exc:
+            raise exc
+
+    @asyncio.coroutine
+    def register(self,
+                 jid,
+                 password,
+                 aux_fields=None,
+                 timeout=60,
+                 no_verify=True
```
horazont: ``no_verify`` must not be set to true by default.

aioxmpp/ibr/xso.py
```diff
@@ -0,0 +1,173 @@
...
+"""
+XSO Definitions
+===============
+
+.. autoclass:: Query
+"""
+
+
+@aioxmpp.IQ.as_payload_class
+class Query(xso.XSO):
```
horazont: This lacks the ``instructions`` field.

aioxmpp/ibr/xso.py
```diff
@@ -0,0 +1,173 @@
...
+    key = xso.ChildText(
+        (namespaces.xep0077_in_band, "key"),
+        default=None,
+    )
+
+    registered = xso.ChildFlag(
+        (namespaces.xep0077_in_band, "registered")
+    )
+
+    remove = xso.ChildFlag(
```
horazont: :+1:

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,253 @@
...
+            type_=aioxmpp.IQType.SET,
+            payload=xso.Query()
+        )
+
+        iq.payload.username = client.local_jid.localpart
+        iq.payload.password = new_pass
+
+        if aux_fields is not None:
+            for key, value in aux_fields.items():
+                setattr(iq.payload, key, value)
```
horazont: I don’t think this is needed. XEP-0077 is rather clear that (in the non-data-form flow) only ``username`` and ``password`` need to be given to change the password. Let’s remove this complexity here for now.

setup.py
horazont: Can we get away with ``>=2.0,<5``? ``2.x`` is known to work at least, I presume you tested ``4.0``, so I *guess* that ``3.0`` will be fine too. This allows operating on older systems, too.

tests/ibr/test_service.py
```diff
@@ -0,0 +1,307 @@
...
+                ['username'],
+            )
+
+    def test_ibr_register(self):
+        password = "aaa"
+        aux_fields = {"nick": "romeo's lover"}
+        with unittest.mock.patch(
+                'aioxmpp.ibr.RegistrationService.connect_and_send'
+        ) as mock1:
+            mock1.return_value = self.auxiliar3()
```
horazont: I think you can use a CoroutineMock from aioxmpp.testutils here. If you need an example of that, let me know; it basically operates like a normal mock, except that it emulates a coroutine. Usage with ``patch`` is like: ``unittest.mock.patch(…, new=CoroutineMock())``, otherwise it’s pretty much the same.

tests/ibr/test_service.py
```diff
@@ -0,0 +1,307 @@
...
+from aioxmpp.testutils import (
+    make_connected_client,
+    run_coroutine,
+)
+
+
+TEST_PEER = aioxmpp.JID.fromstr("juliet@capulet.lit/balcony")
+
+
+class TestService(unittest.TestCase):
```
horazont: Good job on following the conventions & patterns used in the other test modules here! :+1:

aioxmpp/ibr/service.py
```diff
@@ -21,27 +21,123 @@
...
+    reply = yield from aioxmpp.protocol.send_and_wait_for(xmlstream,
+                                                          [iq],
+                                                          [aioxmpp.IQ],
+                                                          timeout=timeout)
+    return reply.payload
+
+
+@asyncio.coroutine
+def register(query_xso,
+             xmlstream,
```
horazont: This is really only a nitpick, but I prefer the argument order ``xmlstream, query_xso``. This is in line with how other send-type functions operate (first the thing over which data is sent, then the data, and then optional flags).

aioxmpp/ibr/service.py
```diff
@@ -21,27 +21,123 @@
...
+    Get a list containing the names of the fields that are used in the
+    xso.Query.
+
+    :param payload: Query object o be
+    :type payload: :class:`xso.Query`
+    :return: :attr:`list`
+    """
+    return [a for a in dir(payload)
+            if isinstance(getattr(payload, a), str) and
+            not a.startswith('__')]
```
horazont: Please make this a method of ``Query`` and use the code I proposed [in my other comment](https://github.com/horazont/aioxmpp/pull/211#discussion_r194223775).

(If you have further questions on how that code works, I’ll be happy to answer.)

aioxmpp/ibr/service.py
```diff
@@ -21,27 +21,123 @@
...
+
+
+def get_query_xso(username, password, aux_fields=None):
```
horazont: This should be the constructor of Query. Also, I suggest to use ``**kwargs`` instead of ``aux_fields``, but I don’t have a strong opinion on that one.

aioxmpp/ibr/service.py
```diff
@@ -56,198 +152,66 @@ class RegistrationService:
...
-        iq.payload.username = client.local_jid.localpart
+        iq.payload.username = self.client.local_jid.localpart
         iq.payload.password = new_pass
 
-        if aux_fields is not None:
-            for key, value in aux_fields.items():
-                setattr(iq.payload, key, value)
-        yield from client.send(iq)
+        if old_pass:
+            iq.payload.old_password = old_pass
```
horazont: Old password is not specified for the non-Data-Forms flow.

aioxmpp/ibr/xso.py
```diff
@@ -94,6 +99,11 @@ class Query(xso.XSO):
         default=None,
     )
 
+    old_password = xso.ChildText(
+        (namespaces.xep0077_in_band, "old_password"),
+        default=None,
+    )
```
horazont: This does not exist in XEP-0077.

aioxmpp/node.py
```diff
@@ -265,31 +265,31 @@ def _try_options(options, exceptions,
...
-            exceptions.append(exc)
-            continue
-        except Exception as exc:
-            protocol.send_stream_error_and_close(
-                xmlstream,
-                condition=(namespaces.streams, "undefined-condition"),
-                text=str(exc),
-            )
-            raise
+        if metadata.sasl_providers:
```
horazont: Add a blank line before and change this to:

```python
if not metadata.sasl_providers:
    return transport, xmlstream, features
```

This saves indentation on the block below.

aioxmpp/security_layer.py
```diff
@@ -1448,8 +1448,7 @@ def certificate_verifier_factory():
         sasl_providers.append(
             AnonymousSASLProvider(anonymous)
         )
-
-    if password_provider is not None or anonymous is False:
+    if password_provider is not None:
```
horazont: Re-insert the blank line before this line.

examples/quickstart_ibr.py
```diff
@@ -0,0 +1,87 @@
+import asyncio
```
horazont: ``quickstart_``-prefix is for things from the Quickstart section in the documentation. Call it ``ibr_test.py``, to indicate that this is not a normal example but just runs a sequence of tests.

tests/ibr/test_service.py
```diff
@@ -185,55 +169,57 @@ def test_ibr_cancel_registration(self):
...
-            self.assertEqual(
-                iq.to,
-                TEST_PEER.bare().replace(localpart=None),
+            iq.payload.username = ''
+            mock1.return_value = iq
+            stream = aioxmpp.protocol.XMLStream(
+                to=TEST_PEER.domain,
+                features_future=asyncio.Future()
             )
+            res = run_coroutine(aioxmpp.ibr.get_registration_fields(stream))
```
horazont: Add a test that the result is indeed the IQ response payload:
```python
self.assertIs(iq.payload, res)
```

tests/ibr/test_service.py
```diff
@@ -185,55 +169,57 @@ def test_ibr_cancel_registration(self):
...
         aux_fields = {"nick": "romeo's lover"}
         with unittest.mock.patch(
-                'aioxmpp.ibr.RegistrationService.connect_and_send'
+                'aioxmpp.protocol.send_and_wait_for',
+                new=CoroutineMock()
         ) as mock1:
-            mock1.return_value = self.auxiliar3()
-            res = run_coroutine(
-                self.s.register(TEST_PEER, password, aux_fields)
+            stream = aioxmpp.protocol.XMLStream(
```
horazont: You are not really using the stream here. Replace it with ``unittest.mock.sentinel.stream`` (which is just a dummy object).

This reduces the coupling between the test code and the XML stream implementation.

tests/test_security_layer.py
```diff
@@ -2866,6 +2866,58 @@ def test_anonymous_without_password_provider(self):
             SecurityLayer(),
         )
 
+    def test__not_anonymous_without_password_provider(self):
```
horazont: Change the name to ``test_not_anonymous_without_password_provider`` (note the first underscore).

aioxmpp/ibr/xso.py
javipalanca: Looking at section 13.4.3 "Change Password" it looks like the field could exist, doesn't it?

tests/ibr/test_service.py
Gersiete: I'm using some attribute of the xmlstream inside the method register, specifically the attribute _to.

aioxmpp/ibr/xso.py
```diff
@@ -181,3 +176,28 @@ class Query(xso.XSO):
     remove = xso.ChildFlag(
         (namespaces.xep0077_in_band, "remove")
     )
+
+
+def get_query_xso(username, password, aux_fields=None):
```
horazont: Why didn’t you make this the constructor of the Query class?

aioxmpp/node.py
```diff
@@ -265,31 +265,34 @@ def _try_options(options, exceptions,
...
-            except Exception as exc:
-                protocol.send_stream_error_and_close(
-                    xmlstream,
-                    condition=(namespaces.streams, "undefined-condition"),
-                    text=str(exc),
-                )
-                raise
+
+        if not metadata.sasl_providers:
+            return transport, xmlstream, features
```
horazont: :+1:

aioxmpp/ibr/service.py
```diff
@@ -0,0 +1,191 @@
...
+    :param xmlstream: Specifies the stream connected to the server where
+                      the account will be created.
+    :type xmlstream: :class:`aioxmpp.protocol.XMLStream`
+
+    :param timeout: Maximum time in seconds to wait for an IQ response, or
+                    :data:`None` to disable the timeout.
+    :type timeout: :class:`~numbers.Real` or :data:`None`
+    """
+    iq = aioxmpp.IQ(
+        to=aioxmpp.JID.fromstr(xmlstream._to),
```
horazont: I’m not 100% happy with the use of the private attribute here. But fixing that is out-of-scope for this PR, so we’ll leave it at this.

tests/ibr/test_service.py
```diff
@@ -0,0 +1,262 @@
...
+            iq_res = aioxmpp.IQ(
+                type_=aioxmpp.IQType.GET,
+                payload=ibr_xso.Query()
+            )
+            iq_res.payload.username = ''
+            mock1.return_value = iq_res
+            stream = aioxmpp.protocol.XMLStream(
+                to=TEST_PEER.domain,
+                features_future=asyncio.Future()
+            )
```
horazont: Instead of creating the stream here, use a mock object:

```python
stream = unittest.mock.Mock(spec=aioxmpp.protocol.XMLStream)
stream._to = TEST_PEER.domain
```

tests/ibr/test_service.py
```diff
@@ -0,0 +1,262 @@
...
+        password = "aaa"
+        aux_fields = {"nick": "romeo's lover"}
+        with unittest.mock.patch(
+                'aioxmpp.protocol.send_and_wait_for',
+                new=CoroutineMock()
+        ) as mock1:
+            stream = aioxmpp.protocol.XMLStream(
+                to=TEST_PEER.domain,
+                features_future=asyncio.Future()
+            )
```
horazont: Same as above.

tests/ibr/test_service.py
horazont: You’re right. I made a new suggestion in the new review.

horazont:
I think I’m happy with this now :). I’ll give @sebastianriese a chance to take another look at it, but from what I can tell it’s great!

(I already used the IBR implementation to register a new test account when my standard test accounts were offline due to a hoster outage :).)

PR #223, opened 2018-07-06
muc: Always strip the resource part of the direct_jid of a member

### LINKED ISSUE #222: XEP-0045: strip resource off direct_jid property of Occupants
The direct_jid is supposed to be always a bare JID.

sebastianriese:
Fixes #222.

### DIFF
--- a/tests/muc/test_service.py
+++ b/tests/muc/test_service.py
@@ -102,7 +102,7 @@ class TestOccupant(unittest.TestCase):
             presence_status=status,
             affiliation="admin",
             role="moderator",
-            jid=TEST_ENTITY_JID
+            jid=TEST_ENTITY_JID.bare()
         )
 
         self.assertEqual(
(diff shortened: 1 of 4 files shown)


tests/muc/test_service.py
```diff
@@ -138,14 +138,30 @@ def test_init_full(self):
...
         )
 
         self.assertEqual(
             occ.is_self,
             unittest.mock.sentinel.is_self,
         )
 
+    def test_init_full_jid_raises_ValueError(self):
+        with self.assertRaisesRegex(ValueError,
+                                    "^the jid argument must be a bare JID$"):
```
horazont: I suggest to use ``r""`` instead of ``""`` for string literals which are used as regex – just so that somebody who modifies it down the road isn’t bitten by it.

sebastianriese: Good point.

PR #224, opened 2018-07-08
Implement support for XEP-0363: HTTP Upload

horazont:
This intentionally does not provide the actual HTTP upload feature, just the XSOs to obtain the slot.

We could add a service to discover the HTTP upload service, but I feel that we might want to put such functionality more generically into DiscoClient or the disco library.


aioxmpp/httpupload/xso.py
```diff
@@ -0,0 +1,175 @@
...
+    .. attribute:: filename
+        :annotation: : str
+
+        The file name (without path, but possibly with "extension") of the
+        file to upload. The server MAY use this in the URL.
+
+
+    .. attribute:: size
+        :annotation: : int
+
```
sebastianriese: Are the ``must``s here intentionally not ``MUST``s?

PR #247, opened 2018-10-31
WIP: Implement in-band bytestreams (XEP-0047)

sebastianriese:
Still missing:
* more complete tests (the `<message>` based flow is not tested at all, perhaps we should simply remove it, and not support it, given that the XEP says, that the IQ-based method SHOULD be used)
* proper end-to-end testing


sebastianriese:
Meh. The tests don't pass on python 3.4, since they use `b"%d" % n`, which was added in 3.5.

horazont:
I wonder whether we should expose something which implements [ReadTransport and WriteTransport](https://docs.python.org/3/library/asyncio-protocol.html) subinterface.

sebastianriese:
Yup, I have thought about it, but not yet investigated in detail. This would indeed be nice for a coherent presentation of the different jingle transports. But as far as I can see, transports can only pass their data on to asyncio protocols, which is clumsy for, let's say, just writing the data to a file. But having the possibility to wrap the stream in a transport may be a good thing.

sebastianriese:
I just read in the asyncio source code <https://github.com/python/cpython/blob/master/Lib/asyncio/transports.py#L233>, and I perhaps should adapt the flow control, to be more similar to the standard behaviour:
* they insulate the transport from exceptions raised by `pause_writing` and `resume_writing`
* when a new buffer limit is set, `pause_writing` is immediately called when appropriate (interestingly enough, `resume_writing` is *not* immediately called)
* the setting of the low/high buffer size limits could behave in the same way.

aioxmpp/ibb/service.py
```diff
@@ -0,0 +1,507 @@
...
+
+from . import xso as ibb_xso
+
+
+MAX_BLOCK_SIZE = (1<<16) - 1
+
+
+def _generate_session_id():
+    # XXX: should we use secrets.token
+    return base64.urlsafe_b64encode(os.urandom(12)).decode("us-ascii")
```
horazont: As discussed, we probably want to refactor the stanza ID generator function into utils and use that instead.

aioxmpp/ibb/service.py
```diff
@@ -0,0 +1,507 @@
...
+
+
+class IBBTransport(asyncio.Transport):
+    """
+    The transport for IBB sessions.
+
+    .. note:: Never instantiate this class directly, all instances of
+              this class are created by :class:`~aioxmpp.ibb.IBBService`.
```
horazont: Maybe add a link to the method(s) which can be used to create such transports?

aioxmpp/ibb/service.py
```diff
@@ -0,0 +1,507 @@
...
+    def get_protocol(self):
+        return self._protocol
+
+    def pause_reading(self):
+        self._reading_paused = True
+
+    def resume_reading(self):
+        self._reading_paused = False
+        if self._protocol is not None:
+            self._protocol.data_received(b"".join(self._input_buffer))
```
horazont: Don’t you need to clear the input buffer here, too?

aioxmpp/ibb/service.py
```diff
@@ -0,0 +1,507 @@
...
+            if len(self._write_buffer) < self._output_buffer_limit_low:
+                if self._protocol is not None:
+                    self._protocol.resume_writing()
+
+            if not self._write_buffer:
+                if self._closing:
+                    break
+                self._can_write.clear()
+
+        close = ibb_xso.Close()
```
horazont: I think you thought this through, but from a quick glance it’s not clear to me that this is sufficiently safe. Don’t we always want to send a Close, even if the write task fails with an exception? What is the suggested flow in the XEP here?

aioxmpp/ibb/service.py
```diff
@@ -0,0 +1,507 @@
...
+            raise errors.XMPPCancelError(
+                condition=errors.ErrorCondition.UNEXPECTED_REQUEST
+            )
+        self._incoming_seq += 1
+        self._data_received(payload.content)
+
+    def _process_msg(self, payload):
+        if payload.seq != self._incoming_seq:
+            return
+        self._incoming_seq += 1
```
horazont: Does the XEP specify a maximum value (and a wraparound) for the sequence number like in SM?

aioxmpp/ibb/service.py
```diff
@@ -0,0 +1,507 @@
...
+        self._data_received(payload.content)
+
+
+class IBBService(service.Service):
+    """
+    A service implementing in-band bytestreams.
+
+    .. attribute:: SESSION_LIMIT = 0
```
horazont: Don’t use ALL_CAPS for non-globals. The ``SESSION_LIMIT`` is an instance attribute, not a class attribute.

aioxmpp/ibb/service.py
```diff
@@ -0,0 +1,507 @@
...
+
+       The maximal number of sessions to be accepted. If there are
+       that many or more active sessions, no new sessions are
+       accepted, unless they are whitelisted by
+       :attr:`expect_session`. (This means, that by default only
```
horazont: ``s/attr/meth/``

aioxmpp/ibb/service.py
```diff
@@ -0,0 +1,507 @@
...
+        return it when it is established. This is meant to be used
+        with signalling protocols like Jingle and is the counterpart
+        to :meth:`open_session`.
+
+        :returns: an awaitable object, whose result is the tuple
+                  `(transport, protocol)`
+        """
+        def on_done(fut):
+            del self._expected_sessions[sid, peer_jid]
+        _, fut = self._expected_sessions[sid, peer_jid] = (
```
horazont: newline between the body of an inline function and the other code pls

aioxmpp/ibb/service.py
sebastianriese: Yes, you are right, the XEP mandates to close the stream when the packet data is erroneous.  My reading skills seem to have been temporarily disabled. This also means we have to keep sessions open, when we received erroneous data (and wait for the close). The only case when we do not send close is when the other side sent a close request or if the other side "vanished". This needs more careful handling of the possible errors.

aioxmpp/ibb/service.py
sebastianriese: Yes it does. It shall wrap as unsigned 16-bit number. (See above comment regarding my reading skills).

aioxmpp/ibb/service.py
horazont: No, what I mean is: after calling ``data_received``, doesn’t it make sense to clear the ``_input_buffer``? Or is that clearing happening elsewhere?

aioxmpp/ibb/service.py
horazont: In case of error we can also just send a close and ignore the response, if it ever arrives. This can be achieved by using ``enqueue`` instead of ``send``.

aioxmpp/ibb/service.py
sebastianriese: the input buffer is cleared in the next line <https://github.com/horazont/aioxmpp/pull/247/files/314cf831ac127d73dbc5330521ffea55d2455c5d#diff-14b03c7eccb3da4229e43ab3f717466dR126>. (The input buffer is a list of `bytes` objects).

horazont: Uff… I’ll steal your statement about reading skills.

aioxmpp/ibb/service.py
```diff
@@ -0,0 +1,568 @@
...
+                        aioxmpp.IQType.SET,
+                        to=self._peer_jid,
+                        payload=ibb_xso.Data(
+                            self._sid,
+                            self._outgoing_seq,
+                            data
+                        )
+                    )
+                elif self._stanza_type == ibb_xso.IBBStanzaType.MESSAGE:
+                    # TODO: use some form of tracking for messages
```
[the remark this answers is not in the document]
sebastianriese: Agreed that the message flow is dangerous.

Issue #30, opened 2016-10-18
Add glossary to the documentation


horazont:
Terms to explain: Basic XMPP terms as well as terms specific to the architecture of aioxmpp.

Issue #26, opened 2016-10-08
Fix sphinx warnings


horazont:
This has had already been fixed in 2bf3596, however, it had to be reverted as my documentation build server cannot deal with the change.

This will be fixed when my doc build server gets updated.

PR #251, opened 2018-11-01
Fix various documentation errors

sebastianriese:
* include api/pulblic/ibr.rst in the toctree
  * make the ibr documentation render properly
* add api/public/version.rst
* add the glossary to the toctree

Note: My sphinx does not seem to be happy with the references generated by `` :xep:`xxx` ``. It emits warnings like
 >  /home/sebi/progs/projects/aioxmpp/docs/user-guide/quickstart.rst:411: WARNING: 4 column based index found. It might be a bug of extensions you use: [('single', 'XMPP Extension Protocols (XEPs); XEP 0092', 'index-3', '')]

### DIFF
--- a/aioxmpp/ibr/service.py
+++ b/aioxmpp/ibr/service.py
@@ -72,7 +72,7 @@ def register(xmlstream,
     Create a new account on the server.
 
     :param query_xso: XSO with the information needed for the registration.
-    :type query_xso: :class:`xso.Query`
+    :type query_xso: :class:`~aioxmpp.ibr.Query`
 
     :param xmlstream: Specifies the stream connected to the server where
                       the account will be created.
@@ -101,7 +101,7 @@ def get_used_fields(payload):
     xso.Query.
 
     :param payload: Query object o be
-    :type payload: :class:`xso.Query`
+    :type payload: :class:`~aioxmpp.ibr.Query`
     :return: :attr:`list`
     """
     return [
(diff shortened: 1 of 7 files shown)


aioxmpp/ibr/service.py
```diff
@@ -135,7 +134,7 @@ def get_client_info(self):
         A query is sent to the server to obtain the client's data stored at the
         server.
 
-        :return: :class:`xso.Query`
+        :return: :class:`~aioxmmp.ibr.Query`
```
horazont: typo

horazont:
See #26 for the warnings. It’s a known issue, but we can’t fix it until the automated doc builder is upgraded.

aioxmpp/ibr/service.py
sebastianriese: I don't see the typo. (Although the text should be reworded.)

horazont: aioxmmp ≠ aioxmpp

horazont:
Re glossary: #30.

aioxmpp/ibr/service.py
sebastianriese: Ow, reading error correction is evil!

sebastianriese:
Re the glossary: There already is some glossary *and* there are links to the glossary throughout the docs. So I would argue in favour of including the glossary in the docs, albeit it may be incomplete. Certainly, all linked glossary entries, that are not yet present, should be added.

Otherwise, if the glossary should be completed first, what is the acceptable minimum (and how well and consistently should the glossary be used throughout the docs?).

I guess, we should definitely explain all of our nomenclature, such as:
* Service
* XSO
* Conversation
* how we use descriptors to register features
and link to the parts of the documentation where those concepts are elaborated.

horazont:
I agree. I think this PR can be merged after you fixed the typo, though.

sebastianriese:
I slightly extended the glossary to include the terms I saw warnings about. I think #30 should remain open nevertheless, since the glossary is still somewhat deficient (also: we have to think about linking discipline, some terms are very common, such as JID and entity, and I think not all occurrences should link to the glossary).

docs/glossary.rst
```diff
@@ -11,10 +11,36 @@ This section defines terms used throughout the :mod:`aioxmpp` documentation.
...
+     :class:`aioxmpp.im.conversation.AbstractConversationMember`.
+
+   Conversation Service
+     A service implementing
+     :class:`aioxmpp.im.conversation.AbstractConversationService`.
+     This allows to create and manage :term:`conversations <Conversation>`.
+
+   Entity
+     An endpoint in the Jabber network, anything that can be addressed by
+     a :term:`JID`. (Compare :rfc:`3920` section 3.1)
```
horazont: Please don’t quote RFC 3920, it is obsoleted by RFC 6120 and RFC 6122. You might want to link to https://tools.ietf.org/html/rfc7622#section-3.1 or https://tools.ietf.org/html/rfc6122 instead.

PR #259, opened 2018-11-11
presence: draft interface for Directed Presence management

### LINKED ISSUE #212: RFC 6121: Support for managing Directed Presence in PresenceServer
For the purpose of this document, **Directed Presence** is a ``<presence/>`` stanza sent to any JID explicitly (i.e. not to the account for the server to distribute). [RFC 6121 § 4.6](https://tools.ietf.org/html/rfc6121#section-4.6) specifies Directed Presence in the XMPP protocol.

Currently, aioxmpp allows sending for Directed Presence (obviously, because arbitrary stanzas can be sent), but it doesn’t support proper management. The reason why additional management is useful/required is that in contrast to normal presence, the server does *not* automatically forward presence updates to entities the user has sent Directed Presence to (unless the entity is already subscribed to the presence).

The main implication is that MUCs to which the user is joined do *not* get presence updates (because they are never subscribed to the users presence) when the users presence changes after having joined.

The PresenceServer should provide an interface to:

1. Send directed presence to an entity (similar to set_presence)
2. Send directed presence to an entity and temporarily subscribe it to presence updates (by automatically delivering presence sent via ``set_presence`` or ``resend_presence`` to that entity). This should be handled by an option to the method provided in (1).
3. When unavailable directed presence to an entity, it should clear the subscription if created per (2).

**up-for-grabs note:** Please see the README and join our chat room if you want help or guidelines on how to implement this.

### LINKED ISSUE #213: XEP-0045: Ensure that presence updates are routed to joined MUCs
This depends on #212.

horazont:
TODO:

* [x] Draft interface
* [ ] Discuss interface
* [ ] Implement
* [ ] Use with `aioxmpp.muc`

Fixes #213, fixes #212.


horazont:
I’d like to get some feedback on the interface before I go about implementing it. Specifically:

* does this likely cover the MUC use cases?
* does this likely cover other use cases?
* will this be able to interact sanely with entitycaps?

aioxmpp/presence/service.py
```diff
@@ -438,3 +591,66 @@ def resend_presence(self):
 
         if self.client.established:
             return self.client.enqueue(self.make_stanza())
+
+    def subscribe_peer_directed(self,
+                                peer: aioxmpp.JID,
+                                muted: bool = False):
+        """
```
sebastianriese: I believe this should take an initial presence as optional argument (so that you can already send a modified directed presence with the request).

aioxmpp/presence/service.py
```diff
@@ -438,3 +591,66 @@ def resend_presence(self):
...
+        creating it with `muted` set to true is the only way to achieve this
+        before the initial directed presence to the peer is sent.
+
+        The newly created handle is returned.
+        """
+
+    def rebind_directed_presence(self,
+                                 relationship: DirectedPresenceHandle,
+                                 new_peer: aioxmpp.JID):
+        """
```
sebastianriese: When is the rebinding necessary? Do we direct presence to full JIDs? (Note: This is probably just my ignorance.)

horazont: Yes, we do for MUC. In MUC, you send directed presence to $bare_muc_jid/$nickname. When changing the nickname (either by choice or because the server forces you), you have to redirect the directed presence. Ugly, but that’s how it is.

aioxmpp/presence/service.py
horazont: That’s what ``muted=True`` is for.

sebastianriese: My argument was less about the capability, but more about convenience of use, and conciseness of the code.

horazont:
> We have to think about the outbound presence filters, I think the directed presence should be in filter chain, perhaps we want an extra chain for directed presence to make it easier for services to work selectively.

The directed presences pass through the outbound presence filter chain like any other presence. It is easy to detect, because it has a non-empty ``@from`` and a type of ``PresenceType.AVAILABLE`` or ``PresenceType.UNAVAILABLE``.

I don’t think we need a specific filter chain for that, especially since the outbound/inbound stanza filters are on the StanzaStream level and the StanzaStream should not need to know the difference.

sebastianriese:
Is directed presence with other types disallowed? Someone may need the capability to send a "busy" presence to their obnoxious relatives ;). I agree it is easy to detect, but I like pre-filtering if it deduplicates code and reduces the number of filter chain calls (by reducing the number of subscriptions).

horazont:
A ``"busy"``` presence is type ``AVAILABLE`` with a specific ``<show/>`` child.

The other presence types are subscription related and not handled by PresenceServer.

aioxmpp/presence/service.py
horazont: Right, but then I’d prefer to be able to set a ``presence_filter`` via the interface.

horazont:
@sebastianriese Mind to give this a review?

Somehow, the tight coupling between the DirectedPresenceHandle and the PresenceServer feels ugly. And also I assume that I’m missing some stuff here because I implemented this over the course of a few weeks with loooong breaks.

aioxmpp/presence/service.py
```diff
@@ -236,6 +236,183 @@ def handle_presence(self, st):
                 self.on_changed(st.from_, st)
 
 
+class DirectedPresenceHandle:
+    """
+    Represent a directed presence relationship with a peer.
+
+    Directed Presence is specified in :rfc:`6121` section 4.6. Since the users
+    server is not responsible for distributing presence updates to peers to
+    which the client has sent directed presence, special handling is needed.
```
sebastianriese: Perhaps we need even more special handling when sending directed presence to peers with from or both subscription (as the RFC specifies that those still receive presence broadcasts). So the current code sends the presence twice to those. (Probably we simply do not want to use this service with those we share subscription with due to the different processing by the server. This should be documented.)

aioxmpp/presence/service.py
```diff
@@ -236,6 +236,183 @@ def handle_presence(self, st):
...
+        Destroy the directed presence relationship.
+
+        The presence relationship becomes useless afterwards. Any additional
+        calls to the methods will result in :class:`RuntimeError`. No additional
+        updates wil be sent to the peer automatically (except, of course, if
+        a new relationship is created).
+
+        If the presence relationship is still active when the method is called,
+        :attr:`~aioxmpp.PresenceType.UNAVAILABLE` presence is sent to the peer
+        immediately. Otherwise, no stanza is sent. The stanza is passed through
```
sebastianriese: Do we want this to happen in all cases? (e.g. to peers that are subscribed to our presence).

aioxmpp/presence/service.py
```diff
@@ -236,6 +236,183 @@ def handle_presence(self, st):
...
+        The *muted* behaviour is useful if presence updates need to be managed
+        by a service for some reason.
+
+        This attribute is read-only. It must be modified through
+        :meth:`set_muted`.
+        """
+        return self._muted
+
+    @property
+    def presence_filter(self):
```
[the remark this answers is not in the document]
horazont: Kind of agreed on the nomenclature. Suggestions?

It is not a ``Filter`` object because the handle is under full control of a single object usually (in contrast to the filters on the stanza stream). If the single object needs to delegate control to multiple other things, it can do that by assigning a filter to that attribute.

aioxmpp/presence/service.py
horazont: We indeed need to document that care should be taken. It’s not as trivial as that, for example when stuff like XEP-0186 (Invisible Command) comes into play.

PR #266, opened 2018-12-17
Allow IQ handlers to respond early

sebastianriese:
This PR does the following things:
* Allow handlers to take additional keyword arguments.
* Use this to add an with_send_reply flag that can be given to IQ handler and then passes an
  as extra argument a function that can be used to send IQ responses *before* the handler function
  terminates. The automatic response will be suppressed by this.

### DIFF
--- a/aioxmpp/service.py
+++ b/aioxmpp/service.py
@@ -729,7 +729,7 @@ class Service(metaclass=Meta):
                         self.__client.stream,
                         obj.__get__(self, type(self)),
                         *additional_args,
-                        **kwargs,
+                        **kwargs
                     )
                 )
 
--- a/aioxmpp/service.py
+++ b/aioxmpp/service.py
@@ -886,7 +886,7 @@ def add_handler_spec(f, handler_spec, *, kwargs=None):
     :type kwargs: :class:`dict`
 
     :raises ValueError: if the handler was registered with
-       different `kwarg`s before
+       different `kwargs` before
 
     This uses a private attribute, whose exact name is an implementation
     detail. The `handler_spec` is stored in a :class:`dict` bound to the
--- a/aioxmpp/service.py
+++ b/aioxmpp/service.py
@@ -210,7 +210,7 @@ import aioxmpp.stream
 
 def automake_magic_attr(obj):
     obj._aioxmpp_service_handlers = getattr(
-        obj, "_aioxmpp_service_handlers", set()
+        obj, "_aioxmpp_service_handlers", dict()
     )
     return obj._aioxmpp_service_handlers
 
@@ -587,7 +587,7 @@ class Meta(abc.ABCMeta):
                     key = next(iter(conflicting))
                     obj = next(iter(
                         obj
-                        for obj_key, obj in SERVICE_HANDLERS
+                        for obj_key, obj, _ in SERVICE_HANDLERS
                         if obj_key == key
                     ))
 
@@ -602,7 +602,7 @@ class Meta(abc.ABCMeta):
 
                 existing_handlers |= unique_handlers
 
-                for spec in new_handlers:
+                for spec, kwargs in new_handlers.items():
                     missing = spec.require_deps - namespace["ORDER_AFTER"]
                     if missing:
                         raise TypeError(
@@ -613,7 +613,7 @@ class Meta(abc.ABCMeta):
                         )
 
                     SERVICE_HANDLERS.append(
-                        (spec.key, attr_value)
+                        (spec.key, attr_value, kwargs)
                     )
 
             elif isinstance(attr_value, Descriptor):
@@ -722,13 +722,14 @@ class Service(metaclass=Meta):
             if isinstance(item, Descriptor):
                 item.add_to_stack(self, self.__context)
             else:
-                (handler_cm, additional_args), obj = item
+                (handler_cm, additional_args), obj, kwargs = item
                 self.__context.enter_context(
                     handler_cm(
                         self,
                         self.__client.stream,
                         obj.__get__(self, type(self)),
-                        *additional_args
+                        *additional_args,
+                        **kwargs,
                     )
                 )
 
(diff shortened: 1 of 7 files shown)


aioxmpp/service.py
```diff
@@ -210,7 +210,7 @@ def handle_on_info_changed(self):
 
 def automake_magic_attr(obj):
     obj._aioxmpp_service_handlers = getattr(
-        obj, "_aioxmpp_service_handlers", set()
+        obj, "_aioxmpp_service_handlers", dict()
```
horazont: ``{}`` instead of ``dict()``

aioxmpp/service.py
```diff
@@ -854,29 +855,74 @@ class HandlerSpec(collections.namedtuple(
...
+       specified keyword arguments
 
     This uses a private attribute, whose exact name is an implementation
-    detail. The `handler_spec` is stored in a :class:`set` bound to the
+    detail. The `handler_spec` is stored in a :class:`dict` bound to the
     attribute.
+
+    The additional `kwargs` arguments must be keyword-only arguments of
+    the function. These args are reduced: only non-default arguments are
+    passed on, this allows robust comparison of additional arguements for
```
horazont: typo (``arguements``)

aioxmpp/service.py
```diff
@@ -854,29 +855,74 @@ class HandlerSpec(collections.namedtuple(
...
+    the comptaibility checking.
```
horazont: Hmmm… Can we get around this by not making the ``with_send_reply`` argument of the ``_apply_iq_handler`` thing not a keyword argument?

Then it would *always* have to be present in the kwargs dictionary and we don’t need this normalization.

(I prefer this way a lot actually, because the ``_reduce_handler_kwargs`` thing feels dangerously close to implementaiton details to me – although it probably isn’t.)

horazont: Also, add a versionchanged note.

aioxmpp/service.py
```diff
@@ -1358,20 +1414,32 @@ def decorator(f):
     return decorator
 
 
-def is_iq_handler(type_, payload_cls, coro):
+def is_iq_handler(type_, payload_cls, coro, kwargs=None):
```
horazont: ``is_iq_handler`` should have the possible kwargs explicit, like:

```python
def is_iq_handler(type_, payload_cls, core, *, with_send_reply=False):
```

and always check them completely. It is used for unit-testing purposes, and since the argument changes the interface, it needs to be checked in all cases.

aioxmpp/stream.py
```diff
@@ -927,31 +927,46 @@ def _destroy_stream_state(self, exc):
...
-            response = request.make_reply(type_=structs.IQType.ERROR)
-            response.error = stanza.Error.from_exception(err)
+            self._send_iq_reply(request, err)
         except Exception:
             response = request.make_reply(type_=structs.IQType.ERROR)
             response.error = stanza.Error(
                 condition=errors.ErrorCondition.UNDEFINED_CONDITION,
                 type_=structs.ErrorType.CANCEL,
             )
+            self._enqueue(response)
```
horazont: How about re-using ``_send_iq_reply`` here (with the created ``stanza.Error`` object)?

aioxmpp/stream.py
```diff
@@ -927,31 +927,46 @@ def _destroy_stream_state(self, exc):
...
             self._logger.exception("IQ request coroutine failed")
         else:
-            response = request.make_reply(type_=structs.IQType.RESULT)
-            response.payload = payload
-        self._enqueue(response)
+            self._send_iq_reply(request, payload)
```
horazont: I’m wondering, we might actually want to split the composition of the IQ reply off of ``_send_iq_reply`` (into ``_compose_iq_reply``). Because composition can fail, and we must still send an IQ error (with undefined-condition) back in that case.

To distinguish cases where composition fails (when we can send undefined-condition back) and where enqueueing fails (where it is not likely that we can do anything useful), we need to split the function.

aioxmpp/stream.py
```diff
@@ -927,31 +927,46 @@ def _destroy_stream_state(self, exc):
...
+
+    def _iq_request_coro_done_check(self, task):
+        try:
+            task.result()
+        except Exception:
+            self._logger.exception("IQ request coroutine failed")
```
horazont: Raising an XMPP error is not "failure" per se. Just a proper way to signal an error.

aioxmpp/stream.py
```diff
@@ -1409,6 +1445,15 @@ def register_iq_request_handler(self, type_, payload_cls, cb):
...
             prefer coroutine functions.
 
+        .. versionadded:: 0.11
+
+            When the argument `with_send_reply` is true `cb` will be
+            called with two arguments: the IQ stanza to handle and a
+            unary function `send_reply(result=None)` that sends a
+            response to the IQ request and prevents that an automatic
+            response is sent. If `result` is an instance of
+            :class:`~aioxmpp.XMPPError` an error result is generated.
```
horazont: Add something like:

> This is useful when the handler function needs to execute actions which happen after the IQ result has been sent, for example, sending other stanzas.

aioxmpp/stream.py
```diff
@@ -2633,18 +2678,26 @@ def iq_handler(stream, type_, payload_cls, coro):
...
+                            early to `cb`.
+    :type with_send_reply: :class:`bool`
 
     The coroutine is registered when the context is entered and unregistered
     when the context is exited. Running coroutines are not affected by exiting
     the context manager.
 
+    .. versionadded:: 0.11
+
+       The `with_send_reply` argument.
```
horazont: Add a ``see :meth:`aioxmpp.stream.StanzaStream.register_iq_request_handler` ``.

tests/test_stream.py
```diff
@@ -594,6 +594,238 @@ def handle_request(stanza):
...
+        self.assertEqual(structs.IQType.RESULT, response_iq.type_)
+        self.assertIs(response_iq.payload, None)
+
+        with self.assertRaisesRegex(RuntimeError,
+                "^send_reply called after the handler is done$"):
+            extracted_send_result()
+
+        self.stream.stop()
+
+
```
horazont: Only one blank line between methods. (only reporting the first instance per file)

aioxmpp/service.py
sebastianriese: Hmm, that sounds right, as long as the handler specs are created uniformly from the decorators, there is no point in doing the normalization (as they will always include values for all kwargs).

aioxmpp/stream.py
sebastianriese: Then I would have to extend the de-multiplexing to make an error reply for certain payload XSOs passed in (but this should be possible via XSO inspection: if it is registered for error or for payload).

aioxmpp/stream.py
sebastianriese: It is an error if we already sent a reply and let the XMPP error bubble up (and only then will this callback be in place).

aioxmpp/stream.py
horazont: aahh, right, yeah, keep it this way, sorry for the Confusion.

aioxmpp/stream.py
horazont: Ah, I got the flow now, thanks!

aioxmpp/service.py
horazont: Yes.

aioxmpp/service.py
```diff
@@ -874,46 +874,34 @@ def func(self):
...
+
+       The `kwargs` argument. If two handlers with the same spec, but
+       different arguments are registered for one function, an error
+       will be raised. So you should always include all possible
+       arguments, this is the responsibility of the calling decorator.
     """
     handler_dict = automake_magic_attr(f)
-    kwargs = _reduce_handler_kwargs(handler_spec, kwargs)
+    if kwargs is None:
+        kwargs = {}
```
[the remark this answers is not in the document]
sebastianriese: I don't like that idiom too much (I'd love to have `None`-short-circuiting operators in Python), but it would be consistent with the rest of aioxmpp.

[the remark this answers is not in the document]
horazont: I don’t care much.

PR #267, opened 2018-12-17
Make ChildTextMap accept either a tag or an XSO

### LINKED ISSUE #264: There should be a shortcut for ChildTextMap
It would be nice if it were possible to define such descriptions without manually defining an XSO for the child: It will be trivial anyway (otherwise we could not use ChildTextMap without stripping information).

sebastianriese:
ChildTextMap now can be constructed from either an XSO or a tag.
If a tag is given a matching XSO is constructed on the fly.

Merging this PR closes #264.

### DIFF
--- a/aioxmpp/xso/model.py
+++ b/aioxmpp/xso/model.py
@@ -1479,10 +1479,23 @@ class ChildTextMap(ChildValueMap):
     convert the :class:`AbstractTextChild` subclass `xso_type` to and from
     a language-text mapping.
 
+    If instead of an :class:`XSO` a tag is passed (that is, a valid
+    argument to :func:`normalize_tag`) an :class:`AbstractTextChild`
+    instance is created on demand.
+
     For an example, see :class:`.Message`.
     """
 
     def __init__(self, xso_type):
+        if not isinstance(xso_type, XMLStreamClass):
+            tag = normalize_tag(xso_type)
+
+            xso_type = type(
+                "TextChild" + tag[1],
+                (AbstractTextChild,),
+                {"TAG": tag},
+            )
+
         super().__init__(
             xso_types.TextChildMap(xso_type),
             mapping_type=structs.LanguageMap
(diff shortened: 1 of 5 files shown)


aioxmpp/xso/model.py
```diff
@@ -1479,10 +1479,23 @@ class ChildTextMap(ChildValueMap):
...
 
     def __init__(self, xso_type):
+        if not isinstance(xso_type, XMLStreamClass):
+            tag = normalize_tag(xso_type)
+
+            xso_type = type(
+                "TextChild" + tag[1],
+                (AbstractTextChild,),
+                {"TAG": tag},
+            )
```
horazont: Hmm, does this work? I’d expect that this complains about the meta class not matching (because ``AbstractTextChild`` has the ``XMLStreamClass`` metaclass, not ``type`` itself).

aioxmpp/xso/model.py
```diff
@@ -2807,3 +2820,50 @@ def to_xso(self):
         instances of their XSO classes) into XSOs.
         """
         return self.xso_class()
+
+
+# XXX: This was move here from aioxmpp/xso/__init__.py
+# and the tests still reside in test___init__.py
```
horazont: Move the tests to ``test_model.py`` and add a test that ``aioxmpp.xso.AbstractTextChild is aioxmpp.xso.model.AbstractTextChild`` to ``test___init__.py``.

tests/xso/test_model.py
```diff
@@ -6024,31 +6024,79 @@ def tearDown(self):
 
 
 class TestChildTextMap(unittest.TestCase):
-    def test_init(self):
-        base = unittest.mock.Mock()
+    def test_init_cls(self):
+        import aioxmpp.xso.model as model
```
horazont: Huh? Why that?

aioxmpp/xso/model.py
sebastianriese: I think it does, but it may be better to use the proper metaclass instead of type. I will investigate, but since the docs say this is the "dynamic form of class" <https://docs.python.org/3/library/functions.html#type> I would guess, that it correctly determines the metaclass from the bases (in which case this form would be resistant against renaming XMLStreamClass or using a subclass of XMLStreamClass as the metaclass of AbstractTextChild).

tests/xso/test_model.py
sebastianriese: I have no idea :D

aioxmpp/xso/model.py
horazont: Ok, agreed. Let’s keep it this way.

PR #272, opened 2019-01-15
hashes: Implement a service and support hashes-used

sebastianriese:
* Support for the hash-used element (used e.g. by jingle file transfer) in addition to hash

* A service that automatically registeres the disco features corresponding to supported hash functions and has tooling to determine hash functions supported in common with another entity.

### DIFF
--- a/aioxmpp/hashes.py
+++ b/aioxmpp/hashes.py
@@ -27,10 +27,14 @@
 Identifiers (usually called `algo`) are defined to refer to specific
 implementations and parametrisations of hashes (:func:`hash_from_algo`,
 :func:`algo_of_hash`) and there is a defined XML format for carrying hash
-digests (:class:`Hash`).
+digests (:class:`Hash`) and hash algorithms to be used (:class:`HashUsed`).
 
 This allows other extensions to easily embed hash digests in their protocols
-(:class:`HashesParent`).
+(:class:`HashesParent`, :class:`HashesUsedParent`).
+
+The service :class:`HashService` registeres the disco features for the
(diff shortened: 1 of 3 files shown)


aioxmpp/hashes.py
```diff
@@ -256,8 +311,92 @@ class HashesParent(xso.XSO):
...
+            self._disco_server.register_feature(feature)
+
+    @asyncio.coroutine
+    def _shutdown(self):
+        for feature in SUPPORTED_HASH_FEATURES:
+            self._disco_server.unregister_feature(feature)
+        yield from super()._shutdown()
+
+    @asyncio.coroutine
+    def select_common_hashes(self, other_entity):
```
horazont: I’m not sure if we might want to have this as a function instead (which takes a DiscoClient and a target JID as arguments). This would alleviate the service of the DiscoServer dependency.

Thoughts?

sebastianriese: As a coroutine function in module scope? That sounds like a good idea (although it does make the HashService even more degenerate).

horazont: Modifying state of a dependent service is a good use for a Service class.

sebastianriese: There is an argument against moving the coroutine out of the service class: consistency. If the HashService is not summoned and we determine the common hashes by service discovery, then we will see the correct list of common hashes, while the other would determine no commonly supported hashes – so we should only be able to use this functionality if we also publish our own hash-features.

horazont: Ah, that probably makes sense.

PR #274, opened 2019-01-21
utils: Make namespaces do consistency checks

sebastianriese:
`namespaces` now is an instance of the new class `Namespaces`, it checks that namespaces don't get two names, that a short-hand is not redefined and prohibits deleting a short-hand.

### DIFF
--- a/aioxmpp/utils.py
+++ b/aioxmpp/utils.py
@@ -30,6 +30,13 @@ Miscellaneous utilities used throughout the aioxmpp codebase.
    Collects all the namespaces from the various standards. Each namespace
    is given a shortname and its value is the namespace string.
 
+   .. note:: This is intended to be promoted to the public API in a
+             future release. Third-party users should not assign
+             short-names for their own namespaces here, but instead
+             use a separate instance of :class:`Namespaces`.
+
+.. autoclass:: Namespaces
+
 .. autofunction:: gather_reraise_multi
 
 .. autofunction:: mkdir_exist_ok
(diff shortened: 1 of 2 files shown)


aioxmpp/utils.py
```diff
@@ -56,15 +63,79 @@
...
-namespaces = types.SimpleNamespace()
+
+class Namespaces:
+    """
+    Manage short-hands for XML namespaces.
+
+    Instances of this class may be used to assign mnemonic short-hands
+    to XML namespaces, for example::
+
+        namespaces = Namespaces()
```
horazont: Let’s make this beautiful. Use:

```rst

    .. code-block:: python

        namespaces = Namespaces() 
        namespaces.foo = "urn:example:foo"
        namespaces.bar = "urn:example:bar" 
```

aioxmpp/utils.py
```diff
@@ -56,15 +63,79 @@
...
+           namespace.example_foo = "urn:example:foo"
+
+    2. no short-hand is redefined to point to a different namespace,
+       continuing the example, the following raises
+       :class:`ValueError`::
+
+           namespaces.foo = "urn:example:foo:2"
+
+    3. deleting a short-hand is prohibited, the following raises
+       :class:`TypeError`::
```
horazont: Maybe AttributeError would be more appropriate? That’s what properties do, too:

```python
>>> class Foo:
...     @property
...     def bar(self):
...         return "x"
...
>>> f = Foo()
>>> del f.bar
  File "<stdin>", line 1, in <module>
AttributeError: can't delete attribute
```

tests/test_utils.py
```diff
@@ -41,6 +41,27 @@ def test_aioxmpp(self):
             "https://zombofant.net/xmlns/aioxmpp#internal"
         )
 
+    def test_inconsistent_rebinding_errors(self):
```
horazont: Those tests should be on a separate TestCase subclass for the ``Namespaces`` class. A separate test here would check that the utils.namespaces object is an instance of ``Namespaces``.

aioxmpp/utils.py
sebastianriese: Probably, I checked a built-in (which raises TypeError), but the analogy to properties is more appropriate.

Issue #269, opened 2019-01-12
XEP-0045: Use stable_id feature to disable body-level matches


horazont:
XEP-0045 version 1.31 provides a feature to indicate that the service will preserve the stanza ID. We could use the presence of this feature to disable costly message body matches altogether.

This requires fetching disco#info of MUCs on join.

PR #276, opened 2019-01-27
muc: self-ping implementation

### LINKED ISSUE #90: XEP-0410, XEP-0045: MUC Self-Ping (Schrödinger's Chat)
Sometimes, MUC services malfunction, there is an s2s outage or other weird issue which gets a client kicked out of a MUC without a ``<presence type="unavailable"/>`` reaching the client.

The usual workaround is to ping oneself. There are a lot of intricacies about this, since the routing of IQ requests is undefined in the case of Multi-Session Nicks (i.e. multiple clients using the same nickname in the same room). This implies fun stuff such as:

* We have to use high timeouts because the entity which eventually receives the ping request may be a mobile client whose session is currently in hibernation and it’ll only reply to the Ping in five minutes when it’s back in coverage.
* The ping may be forwarded to an entity which doesn’t even support the XEP-0199 (XMPP Ping) protocol, so we get a ``<feature-not-implemented/>`` or ``<service-unavailable/>`` back instead of a proper pong.

Depends on #77

horazont:
Fixes #90.

- [x] Implement business logic of XEP-0410 in separate classes
- [x] Inhibit sending pings (but do not change state otherwise) during SM hibernation.
- [x] Ensure that cancelling a send() operation on an IQ removes all data associated with the IQ from the StanzaStream
- [x] Ensure that cancelling a ping() operation properly propagates the cancel to the send()
- [x] Make Room use those classes
- [x] Add end-to-end tests


horazont:
In the meantime, we might discuss the default timeouts. The current code sets:

- 12 minutes soft-timeout. When the soft-timeout trips, the pinger is started.
- 15 minutes hard-timeout. When the hard-timeout trips, on_muc_stale is emitted, notifying the application that the MUC might be unreachable.
- 2 minutes ping interval. Every 2 minutes (while the pinger is running), a ping is sent.
- 8 minutes ping timeout. A ping can take up to 8 minutes to be replied to before it is counted as timed-out. This means that up to 4 pings are in-flight at the same time.

horazont:
> Hmmm. The default time-outs seem very high to me. 

The hard/softlimits are high, because this failure mode should be rare, and we shouldn’t be wasting bandwidth and stuff on it.

The ping timeout is high because in a MUC without explicit XEP-0410 support, we might be pinging a mobile client, which may take a while to reply for network reasons.

The ping interval is high to not have too many pings in-flight with a high ping timeout.

> We might want to scale them with the last known number of members in the MUC (so that the clients don't collectively bother large, quiet MUCs with a ping-flood, but get faster disconnect detection for small, quiet MUCs).

I don’t see this as a problem at the moment. I thought of adding a 10% jitter or something to the MUCPinger though.

Scaling them by the number of occupants seems rather arbitrary for me (once you take the load on the MUC service out of the equation).

> The current time-outs to not seem to work together well. Two pings are sent before the hard timeout trips, and the first one of those has less than half its default life-time to be replied before the MUC is marked stale by the hard timeout.

There’s a trade-off to be made here. In ideal conditions, we should be seeing an "immediate" reply to the ping. Otherwise, stuff *is* already stale. Of course, if we accidentally ping a mobile client, there are good reasons why a ping even in a non-stale MUC takes a while. This will lead to the user interface flapping between stale and non-stale, unfortunately.

I think we should also adapt the defaults of those timeouts according to whether the MUC has support for XEP-0410 (a disco#info feature for detecting that is in the pipeline). For a MUC which supports this feature, I suggest to set the following timeouts:

```python
hard_timeout = soft_timeout - timedelta(minutes=3)
ping_interval = timedelta(minutes=2)
ping_timeout = timedelta(minutes=1)
```

Ideally, this would scale with the streams (measured) round-trip time, but that’s more tricky to do.

I’d like to postpone the automatic configuration of timeouts to a PR which’ll also address #269 (since we need to know disco#info of the MUC room for that).

aioxmpp/stream.py
```diff
@@ -2655,16 +2656,23 @@ def handler_error(exc):
...
+            if not timeout:
+                reply = yield from fut
+            else:
+                try:
+                    reply = yield from asyncio.wait_for(
+                        fut,
+                        timeout=timeout
+                    )
+                except asyncio.TimeoutError:
+                    raise TimeoutError
```
sebastianriese: Do we want `TimeoutError` instead of `asyncio.TimeoutError` beyond being backwards compatible, I know we use the system level exceptions at other places – but do we really want that.

What does a grep through the code base say? (That is a discussion that must be concluded before aioxmpp 1.0).

aioxmpp/utils.py
```diff
@@ -344,3 +348,217 @@ def to_nmtoken(rand_token):
...
+
+    @deadtime_hard_limit.setter
+    def deadtime_hard_limit(self, value):
+        if self._hard_limit_timer is not None:
+            self._hard_limit_timer.cancel()
+        self._hard_limit = value
+        self._retrigger_timers()
+
+
+class proxy_property:
```
sebastianriese: Why not implement this as a function which provides property with the appropriate functions? Along the lines:
```python
def proxy_property(owner_attr, member_attr, *, readonly=False, allow_delete=False, doc=None):
      getter = lambda self: getattr(getattr(self, owner_attr), member_attr)
      setter = None if readonly else lambda self, value: ...
      ....
      return property(getter, setter, deleter, doc)
```
IMO this is the more readable variant.

One could even think about compiling the lambdas from source to avoid the lookup of getattr for better performance (of course at the cost of worse error detection if the attr strings contain non-indentifiers, so probably not a good idea).

aioxmpp/muc/self_ping.py
```diff
@@ -0,0 +1,429 @@
...
+                self._on_fresh()
+                return
+
+            if exc.condition == aioxmpp.errors.ErrorCondition.ITEM_NOT_FOUND:
+                return
+
+            self._on_exited()
+
+    @asyncio.coroutine
+    def _pinger(self):
```
sebastianriese: The resend timeout should be randomized (uniformly, probably in a range `[0.5..1.5]*base_timeout`) to avoid clustering of pings after a MUC got silent.

aioxmpp/stream.py
horazont: I intentionally chose TimeoutError here. If you want asyncio.TimeoutError, you can implement it yourself with ``asyncio.wait_for`` or something.

aioxmpp/utils.py
horazont: I don’t think the lambdas are more readable, and building source code is even less of an option. I went the "building source code" route with xsltea, and I’d like to never go down that rabbithole again.

sebastianriese: well then don't use lambdas but local functions, then one can even use the decorator form of property – I just somehow prefer reusing property instead of rolling a weird property-like descriptor.

horazont: Using local functions makes it structurally very similar to a class then.

I don’t see value in forcing a re-use of property in this case. This variant can easily be extended to fetch a docstring if a type for the member is given, too (helping with sphinx). I’m not sure how easy that’s possible with a property wrapper.

horazont: I found a reason to re-use ``property``: It provides the same interface to override setters or getters in derived classes—thus making it more transparent in its use.

horazont: Done.

aioxmpp/muc/self_ping.py
horazont: This will screw up the tests baaaad. I have no idea how to do this while still having timing tests which run at least mostly reliable.

Although I see that this is a desired feature. Probably 10% jitter is enough though.

sebastianriese: Yes it probably is (the +- 50% jitter was inspired by RTCP).

aioxmpp/muc/self_ping.py
```diff
@@ -31,6 +32,10 @@
 import aioxmpp.utils
 
 
+def _apply_jitter(v, amplitude):
+    return v + (random.random() * 2 - 1) * amplitude
```
sebastianriese: Should this not be `v * (1 + amplitude * (2*random.random() - 1))` for 10% relative jitter?

horazont: Fixedyfix.

PR #284, opened 2019-03-02
docs: decide whether to adopt Google Docstring style

horazont:
See the diff for an example.

**The only difference is in the source.** The rendered output is identical.

I think this might be much more readable. Shout out to @f3anaro for pointing this out to me.

If we decide to go for this, I suggest we follow the "clean what you’re touching" method instead of making an overhaul of all docstrings. I.e.: when you modify a function, adapt its docstring on the way. Just like we’re still in the process of moving to ``:param:`` instead of prose-only.

### DIFF
--- a/aioxmpp/node.py
+++ b/aioxmpp/node.py
@@ -421,22 +421,20 @@ class Client:
     """
     Base class to implement an XMPP client.
 
-    :param local_jid: Jabber ID to connect as
-    :type local_jid: :class:`~aioxmpp.JID`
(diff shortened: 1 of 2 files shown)


sebastianriese:
Hmm, I don't like adding another layer of docstring-convention transitioning to the codebase, but the style does indeed look much more readable. (The transition should be complete when we reach 1.0.)

horazont:
Ok, since we agree on the readability, I think we should go ahead on this one.

PR #286, opened 2019-03-02
security_layer: support ssl_context_factory argument on make

### LINKED ISSUE #283: make_security_layer should allow passing an ssl context factory
The tls_with_password_based_authentication had it, and it’s needed in some cases such as #282.

horazont:
Fixes #283.

### DIFF
--- a/aioxmpp/security_layer.py
+++ b/aioxmpp/security_layer.py
@@ -1428,15 +1428,25 @@ def make(
 
     .. note::
 
-       :data:`False` and ``""`` are treated differently for the `anonymous`
-       argument, despite both being false-y values!
+        :data:`False` and ``""`` are treated differently for the `anonymous`
+        argument, despite both being false-y values!
+
+    .. note::
+
+        If `anonymous` is not :data:`False` and `password_provider` is not
+        :data:`None`, both authentication types are attempted. Anonymous
+        authentication is, in that case, preferred over password-based
+        authentication.
+
+        If you need to reverse the order, you have to construct your own
+        :class:`SecurityLayer` object.
 
     .. warning::
 
-       Take the security and privacy considerations from :rfc:`4505` (which
-       specifies the ANONYMOUS SASL mechanism) and :xep:`175` (which discusses
-       the ANONYMOUS SASL mechanism in the XMPP context) into account before
-       using `anonymous`.
+        Take the security and privacy considerations from :rfc:`4505` (which
+        specifies the ANONYMOUS SASL mechanism) and :xep:`175` (which discusses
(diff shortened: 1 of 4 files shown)


aioxmpp/security_layer.py
```diff
@@ -1376,6 +1381,12 @@ def make(
     :data:`None` while `pin_store` is not, a coroutine which returns
     :data:`False` is substituted.
 
+    `ssl_context_factory` can be a callable taking no arguments and returning
+    a :class:`OpenSSL.SSL.Context` object. If given, the factory will be used
+    to obtain an SSL context when the stream negotiates transport layer security
+    via TLS. By default, :func:`aioxmpp.security_layer.default_ssl_context` is
+    used, which should be fine for most applications.
```
sebastianriese: This should sound more like the hdparm manual page; we should note that you should only use another context if you know what you are doing as using a misconfigured context can disable the authentication of the peer.

tests/test_security_layer.py
```diff
@@ -2425,6 +2425,52 @@ def test_simple(self):
...
+                )
+            )
+
+            PKIXCertificateVerifier = stack.enter_context(
+                unittest.mock.patch(
+                    "aioxmpp.security_layer.PKIXCertificateVerifier"
+                )
+            )
+
+            default_ssl_context = stack.enter_context(
```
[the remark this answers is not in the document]
horazont: I thought that too (I copied the code from the other test case), but then I realized that there is in fact a semantic difference which can only be caught by mocking the function: if one sets the ``ssl_context_factory=default_ssl_factory`` at function definition time in the parameter list, an application cannot monkey patch ``default_ssl_factory``.

But you’re right that it isn’t even needed in this test case, so it should be dropped.

aioxmpp/security_layer.py
```diff
@@ -1398,22 +1428,36 @@ def make(
...
-       Take the security and privacy considerations from :rfc:`4505` (which
-       specifies the ANONYMOUS SASL mechanism) and :xep:`175` (which discusses
-       the ANONYMOUS SASL mechanism in the XMPP context) into account before
-       using `anonymous`.
+        Take the security and privacy considerations from :rfc:`4505` (which
+        specifies the ANONYMOUS SASL mechanism) and :xep:`175` (which discusses
+        the ANONYMOUS SASL mechanism in the XMPP context) into account before
+        using `anonymous`.
 
     The versaility and simplicity of use of this function make (pun intended)
```
sebastianriese: If you are cleaning up the docs anyways: "vesaility" => "versatility"

horazont: Done!

PR #287, opened 2019-03-10
travis: use archive tarballs instead of full hg clones

horazont:
Saves a lot of download :)


horazont:
Obsolete, as we now install from apt instead.

PR #296, opened 2019-08-14
xml: fix serialisation of re-ocurring nested namespaces

### LINKED ISSUE #295: Missing namespaces in forwarded message
When sending a message with a forwarded message, all containers in forwarded stanza  have no namespaces

horazont:
This fixes an oversight where namespace URIs whose prefix has
been re-bound would not be removed from _curr_ns_map. That leads
to the prefix/namespace not being re-declared if the namespace was
declared further up in the hierarchy, but the prefix used in the
initial declaration was re-bound in-between.

Please see the added test cases for example situations.

Fixes #295.

### DIFF
--- a/aioxmpp/xml.py
+++ b/aioxmpp/xml.py
@@ -312,16 +312,32 @@ class XMPPXMLGenerator:
 
         new_decls = self._ns_decls_floating_in
         new_prefixes = self._ns_prefixes_floating_in
+        old_ns_map = self._curr_ns_map
         self._ns_map_stack.append(
             (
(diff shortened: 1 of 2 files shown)


horazont:
This most certainly needs a backport to 0.10.

horazont:
Backported to 0.10 for release 0.10.5.

horazont:
(Going to ignore the ``push`` travis build since it fails due to bitrot. This PR is based on the common-ancestor of release-0.10 and devel for backporting.)

PR #298, opened 2019-08-19
Fix lxml dependency declaration for Python < 3.5

horazont:
lxml dropped support for Python 3.4 in version 4.4.


horazont:
Backported to 0.10 as 6f2570c0b3b6603c50738334e614e8412225cde8.

Issue #312, opened 2019-09-23
MUC Self-Ping: stack-trace / no return after outage


ge0rg:
The current code is exhibiting a stacktrace and not coming back to a MUC after an outage. It also looks like it tries to leave the "broken" room twice:

```
2019-09-23 01:09:48,232 DEBUG aioxmpp.node.Client.muc.MUCClient.MUCMonitor room@chat.yax.im/roombot: soft-limit tripped, starting pinger
2019-09-23 01:09:48,233 DEBUG aioxmpp.node.Client.muc.MUCClient.MUCMonitor room@chat.yax.im/roombot: request to start pinger
2019-09-23 01:09:48,234 DEBUG aioxmpp.node.Client.muc.MUCClient.MUCMonitor room@chat.yax.im/roombot: request to stop pinger
2019-09-23 01:09:48,235 DEBUG aioxmpp.node.Client.muc.MUCClient.MUCMonitor room@chat.yax.im/roombot: already stopped
2019-09-23 01:09:48,236 DEBUG aioxmpp.node.Client.muc.MUCClient.MUCMonitor room@chat.yax.im/roombot: pinger booted up
2019-09-23 01:09:48,237 DEBUG aioxmpp.node.Client.muc.MUCClient.MUCMonitor room@chat.yax.im/roombot: pinger loop. interval=datetime.timedelta(0, 120)
2019-09-23 01:09:48,238 DEBUG aioxmpp.node.Client.muc.MUCClient.MUCMonitor room@chat.yax.im/roombot: sending self-ping with timeoutdatetime.timedelta(0, 480)
2019-09-23 01:09:48,239 DEBUG aioxmpp.node.Client.muc.MUCClient.MUCMonitor room@chat.yax.im/roombot: pinger waiting for 1 pings forat most 120.0s
2019-09-23 01:09:48,240 DEBUG aioxmpp.StanzaStream sending <iq from=None to='room@chat.yax.im/roombot' id=':f-3N6imBswtvm7Izcvg3' type=<IQType.GET: 'get'> data=<aioxmpp.ping.xso.Ping object at 0xb28f7c8c>> and waiting for it to be sent
[...]

PR #314, opened 2019-09-25
muc: treat remote server unreachability like timeout self-ping

horazont:
If the remote server is unreachable network-wise, we cannot know
whether we have been removed. In addition, a rejoin will fail
deterministically.

It is thus better to stay in the undecided pinging state (treat it
like a timeout): this will allow us to wait for a rejoin until we
know more about the current state and the server is actually
reachable again.

See-Also: https://github.com/xsf/xeps/pull/834
See-Also: #312


horazont:
We should defer the merge until the XEP update has passed by council.

PR #330, opened 2020-04-13
Improvements to the usability of the e2etest suite

horazont:
* Add ``--e2etest-only`` flag to ``python3 -m aioxmpp.e2etest`` to allow skipping all non-e2etests
* Add provisioner which can be configured with a fixed set of accounts.
* Add a Quirk to disable XEP-0049-based tests

### DIFF
--- a/tests/pep/test_e2e.py
+++ b/tests/pep/test_e2e.py
@@ -29,7 +29,7 @@ import aioxmpp.pubsub.xso as pubsub_xso
 from aioxmpp.e2etest import (
     blocking,
     blocking_timed,
-    require_identity,
+    require_pep,
     TestCase
 )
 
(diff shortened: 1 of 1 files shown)


horazont:
@guusdk Please test in this order, so that I can be sure that everything works:

1. Set the new quirk for Private XML to see if all private XML related errors disappear
2. Change the provisioner to the StaticPasswordProvisioner and remove the quirk

aioxmpp/e2etest/provision.py
```diff
@@ -725,3 +728,100 @@ def _make_client(self, logger):
...
+    """
+    This provisioner expects a list of username/password pairs to authenticate
+    against the tested server.
+
+    This is for use with servers which support neither SASL ANONYMOUS nor
+    a ``mod_auth_any`` equivalent.
+
+    The configuration of this provisioner is slightly unwieldly since we do
+    not want to add a dependency to a more sane configuration file format. Here
+    is an example on how to configure a provisioner with three accounts:
```
guusdk, line 743: Three? The example shows two.

horazont: That’s because I got lazy after the first two. Thanks.

PR #346, opened 2020-08-27
test_structs: Add jid escape testcase

lovetox:
Adds a simple test case which fails at the moment

### DIFF
--- a/tests/test_structs.py
+++ b/tests/test_structs.py
@@ -52,6 +52,8 @@ ESCAPING_TEST_VECTORS = [
     (r"\2plus\2is\4", r"\2plus\2is\4"),
     (r"foo\bar", r"foo\bar"),
     (r"foo\41r", r"foo\41r"),
+    # additional test vectors
+    (r"call\20me", r"call\5c20me"),
 ]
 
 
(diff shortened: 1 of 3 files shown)


tests/test_structs.py
```diff
@@ -52,6 +52,7 @@
     (r"\2plus\2is\4", r"\2plus\2is\4"),
     (r"foo\bar", r"foo\bar"),
     (r"foo\41r", r"foo\41r"),
+    (r"call\20me", r"call\5c20me"),
```
horazont, line 55: @lovetox This doesn’t quite fit into the `exceptions` section, maybe put it into a new "additional test vectors" section.

This is also a good candidate for the XEP.

horazont (changes requested):
Thanks and good catch, please fix the minor issue.

horazont:
@lovetox If you want, you can add a changelog entry in the Version 0.12 section [in the changelog](https://github.com/horazont/aioxmpp/blob/devel/docs/api/changelog.rst).

Also, please squash the commits together. I do not want to have a deterministically failing test in case I ever need to bisect one.

If you prefer to have the breakage documented, please use an [expectedFailure](https://docs.python.org/3/library/unittest.html#unittest.expectedFailure) decoration.

PR #365, opened 2021-10-28
Add Python 3.10 to the testing

cclauss:
https://pythoninsider.blogspot.com/2021/10/python-3100-is-available.html

https://docs.python.org/release/3.10.0/whatsnew/changelog.html#python-3-10-0-final

The loop parameter has been removed from most of asyncio‘s high-level API following deprecation in Python 3.8.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -25,6 +25,7 @@ jobs:
         - '3.7'
         - '3.8'
         - '3.9'
+        - '3.10'
         test-type:
         - e2e
         e2e-software:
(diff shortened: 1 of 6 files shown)


horazont:
@cclauss Thanks for the contribution. I expect the 3.10 pipeline to break for multiple reasons (I think my dear nosetests breaks on 3.10, too), so it's good to have it monitored.

cclauss:
Wha about https://github.com/nose-devs/nose2 or https://pypi.org/project/nose3 instead of nose.

horazont:
Woop woop! Thanks a lot for triggering this, now it finally works :)

PR #370, opened 2021-11-20
Fix docstring example from class XSOParser

haxtibal:
Minor adaptions to make the example work when copy pasting it:
- typo: lmxl
- [lxml.sax.saxify](https://lxml.de/api/lxml.sax-module.html#saxify) needs a content handler as 2nd argument
- message must be namespaced to work with referenced `class Message`
- message must have tag "to" to work with referenced `class Message`
- result is seemingly at global scope, so `nonlocal` doesn't work

### DIFF
--- a/aioxmpp/xso/model.py
+++ b/aioxmpp/xso/model.py
@@ -2571,15 +2571,16 @@ class XSOParser:
         # let Message be a XSO class, like in the XSO example
         result = None
         def catch_result(value):
-            nonlocal result
+            global result
             result = value
 
         parser = aioxmpp.xso.XSOParser()
         parser.add_class(Message, catch_result)
         sd = aioxmpp.xso.SAXDriver(parser)
-        lxml.sax.saxify(lmxl.etree.fromstring(
-            "<message id='foo' from='bar' type='chat' />"
-        ))
+        lxml.sax.saxify(lxml.etree.fromstring(
+            "<jc:message id='foo' from='bar' to='baz' type='chat' "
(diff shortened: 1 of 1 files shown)


aioxmpp/xso/model.py
```diff
@@ -2571,15 +2571,16 @@ class XSOParser:
         # let Message be a XSO class, like in the XSO example
         result = None
         def catch_result(value):
-            nonlocal result
+            global result
```
horazont, line 2574: Ah, this is tricky. Often you'd have this code in a function, where `nonlocal` is the correct choice. But on the global scope that doesn't work :/.

I don't care much either way, so let's keep the new one.

aioxmpp/xso/model.py
```diff
@@ -2571,15 +2571,16 @@ class XSOParser:
...
             result = value
 
         parser = aioxmpp.xso.XSOParser()
         parser.add_class(Message, catch_result)
         sd = aioxmpp.xso.SAXDriver(parser)
-        lxml.sax.saxify(lmxl.etree.fromstring(
-            "<message id='foo' from='bar' type='chat' />"
-        ))
+        lxml.sax.saxify(lxml.etree.fromstring(
+            "<jc:message id='foo' from='bar' to='baz' type='chat' "
```
horazont, line 2581: FWIW, this would also have worked with a default namespace instead of a prefixed one.

horazont (commented):
Thanks, merging! :)

PR #371, opened 2021-12-09
Fix typos

kianmeng:


aioxmpp/protocol.py
```diff
@@ -498,7 +498,7 @@ def _rx_feed(self, blob):
         except Exception:
             self._logger.exception(
                 "unexpected exception while parsing stanza"
-                " bubbled up through parser. stream so ded.")
+                " bubbled up through parser. stream so dead.")
```
horazont, line 501: So this was actually deliberate as a kind of meme joke at the time. Probably good riddance though anyway :).

aioxmpp/xmltestutils.py
```diff
@@ -22,7 +22,7 @@
 import unittest
 
 
-def element_path(el, upto=None):
+def element_path(el, up to=None):
```
horazont: This is invalid syntax.

horazont (changes requested):
The syntax error needs to be fixed, otherwise this is great, thank you!

horazont (approved):

horazont:
@kianmeng Thanks a lot!

Issue #136, opened 2017-08-29
Add type annotations all over aioxmpp


horazont:
[PEP 484](https://www.python.org/dev/peps/pep-0484/) introduced a standard for type annotations in Python. There are already checkers and IDEs out there which understand and can greatly benefit from type annotations. The annotations have no runtime cost (and do not provide strict runtime type checking in any way).

The annotations are useful documentation for IDEs and checkers and should be included to improve the accuracy of automated static analysis.

**up-for-grabs note:** Please see the README and join our chat room if you want help or guidelines on how to implement this.

PR #377, opened 2022-01-12
ping: add type annotations

SamWhited:
This adds a few type annotations for the `ping` package. Despite the simplicity of the package I had a few notes and questions:

- I couldn't figure out what the type of the request argument should be
- `PingService.ping` returns the result of the `ping` function, but the `ping` function does not return the results of sending the IQ. Was this intentional or should it be updated and the type changed to an option type?

Updates #136

### DIFF
--- a/aioxmpp/ping/service.py
+++ b/aioxmpp/ping/service.py
@@ -25,6 +25,8 @@ import aioxmpp.disco
 import aioxmpp.service
 import aioxmpp.structs
 
+from aioxmpp.node import Client
+from aioxmpp.stanza import IQ
 from aioxmpp.utils import namespaces
 
 from . import xso as ping_xso
@@ -49,10 +51,10 @@ class PingService(aioxmpp.service.Service):
     )
 
     @aioxmpp.service.iq_handler(aioxmpp.structs.IQType.GET, ping_xso.Ping)
-    async def handle_ping(self, request):
+    async def handle_ping(self, request: IQ) -> ping_xso.Ping:
         return ping_xso.Ping()
 
-    async def ping(self, peer):
+    async def ping(self, peer: structs.JID) -> None:
         """
         Wrapper around :func:`aioxmpp.ping.ping`.
 
@@ -84,7 +86,7 @@ class PingService(aioxmpp.service.Service):
         return await ping(self.client, peer)
 
 
-async def ping(client, peer):
+async def ping(client: Client, peer: JID) -> None:
     """
     Ping a peer.
 


horazont:
Thanks for that. 

As you can see, the CI complains about this a little. The reason is that you cannot import the aioxmpp root package because that would cause a cyclic import: `aioxmpp/__init__.py` imports a bunch of sub-packages to re-export their identifiers.

*Edit:* the workaround/fix for that is to import the types from their original location for annotations, or use string-based type annotations (though I find the latter uglier).

To answer your questions:

> PingService.ping returns the result of the ping function, but the ping function does not return the results of sending the IQ. Was this intentional or should it be updated and the type changed to an option type?

IIRC intentional -- this avoids forgetting to update the ping() function if we ever decide to return anything meaningful from the ping method.

> I couldn't figure out what the type of the request argument should be

That should be `aioxmpp.stanza.IQ`, but I'm not sure if mypy gets that.

SamWhited:
Thanks; I believe I've fixed it.

horazont:
@SamWhited Not yet, I'm afraid. You cannot have `import aioxmpp` or `from aioxmpp import` in any of the subpackages.

SamWhited:
@horazont okay, I guess I'm misunderstanding your original post. It looks like `service.py` already uses `aioxmpp.IQ` (although I don't see how as it's not imported) and `xso.py` seems to import `aioxmpp` already? It's been a number of years since I worked in python though, so sorry for the misunderstanding.

horazont:
`ping/xso.py` does *not* import aioxmpp, but aioxmpp.utils, which does not import aioxmpp itself.

To get the identifiers you need for the annotations, you can use:

```python
from aioxmpp.stanza import IQ
from aioxmpp.node import Client
from aioxmpp.structs import JID
```

if I recall the file names correctly :)

SamWhited:
> ping/xso.py does not import aioxmpp, but aioxmpp.utils, which does not import aioxmpp itself.

https://github.com/horazont/aioxmpp/blob/devel/aioxmpp/ping/xso.py#L22

I fixed this again however using the imports you suggested. Fingers crossed it builds this time!

SamWhited:
Still failing with no useful error; I'm afraid I'm out of ideas. Sorry, thought this would be a quick easy fix to get rid of at least one error from annoying python tools that complain about things I mostly can't fix on my end anyways. Oh well, if I think of anything else I'll push more changes.

horazont:
Ugh, that's a nasty import loop you go there. Sorry, there's probably no trivial fix for this.

PR #378, opened 2022-01-17
typo fix: no_verfiy -> no_verify

sseyren:


utils/ejabberd-cfg/e2etest.ini
```diff
@@ -5,6 +5,6 @@ provisioner=aioxmpp.e2etest.provision.AnonymousProvisioner
 host=localhost
 pin_store=.travis-pinstore.json
 pin_type=0
-no_verfiy=true
+no_verify=true
```
horazont: So all of these have obviously been ineffective and also incorrect semantically; we use a pin store to avoid having to bypass verification. Would you like to add a commit which removes those?

If you do, *make sure* to include the rationale, because I don't want that to be lost.

horazont (commented):
Thanks, just one note inline.

horazont (changes requested):
Hey! Thanks for the update. Please include the information *why* it was ok to remove the `no_verify`. [Context is key](https://cbea.ms/git-commit/) in commit messages, and I (or anyone) doesn't want to stare at the diff in confusion without context half a year down.

sseyren:
Ok, now I think that I wrote a proper commit message for changes. Also, thank you for pointing me details that I need to consider before open a PR for this project.

horazont:
Thanks! Commit message looks great now, let's make a quick pipeline check.

horazont (approved):

